PR #183, opened 2018-01-24
 Make dropdowns scrollable and positioned within the viewport

tomaskikutis:


examples/templates/dropdown.html
```diff
@@ -135,88 +135,6 @@ <h3 class="docs-page__h3">Dropdowns opened on the side of the trigger.</h3>
     </div>
 </div>
 
-
-<h3 class="docs-page__h3">Dropdown Append To Body</h3>
```
pavlovicnemanja: Hey, why you removed completely dropdown append to body section? we need this dropdowns, on nested places where we have overflow scroll or overflow hidden.

tomaskikutis: After this change, all dropdowns are appended to the body by default.

tomaskikutis: Actually, only the ones two levels deep and deeper are appended to the body. I will update the PR tomorrow, so the first one is also appended.

PR #213, opened 2018-05-03
Refactor editor popup

tomaskikutis:
SDFID-390


tomaskikutis:
Most if not all the code from `_sd-comment-box.scss` was renamed and moved to `_sd-editor-popup.scss`. I've also updated the playground so you should be able to reuse it.

PR #226, opened 2018-06-18
feat(lineInput): Add line input directive

pavlovicnemanja:


petrjasek:
imo this would be better done in superdesk-client-core, then we could handle there common error messages using translations

petrjasek:
I think it would be better if the ui framework handles the css and not the components, we should have components in core and not use the ui framework classes in views if possible, then it would be kind of decoupled. not sure if I see a benefit of having the components here, is there any project using the framework and not using client core?

pavlovicnemanja:
Liveblog is also using UI Framework and Publisher is using UI Framework (but it is on superdesk-core)... Also, benefit of having all this in UI framework is that we can create UI prototypes and show them to client.

petrjasek:
@fritzSF I mean that we shouldn't write code like:
```html
<div class="sd-column-box__slide-in-column sd-column-box__slide-in-column--large sd-column-box__slide-in-column--dark-ui">
            <div class="sd-column-box__slide-in-column-inner sd-slide-in-panel">
                <div class="sd-slide-in-panel__header">
                    <h3 class="sd-slide-in-panel__heading" translate>Metadata</h3>
                </div>
                <div class="sd-slide-in-panel__content">
                    <div class="sd-slide-in-panel__content-block sd-slide-in-panel__content-block--first">
                        <div class="sd-alert sd-alert--hollow sd-alert--primary sd-alert--small" style="margin: -1rem 0 3rem;">
                            <span translate>Currently editing: {{images.length}} items</span>
                        </div>
                        <div class="form__row" ng-if="validator.headline">
                            <div class="sd-line-input sd-line-input--dark-ui sd-line-input--boxed" ng-class="{'sd-line-input--required':validator.headline.required}">
                                <label class="sd-line-input__label" translate>Title</label>
                                <input class="sd-line-input__input" type="text" ng-model="metadata.headline" ng-change="onChange('headline')" ng-blur="onBlur()" ng-disabled="!images.length" placeholder="{{placeholder['headline']}}">
                                <div class="sd-line-input__char-count" ng-if="validator.headline.maxlength" sd-character-count data-item="metadata.headline" data-limit="validator.headline.maxlength" data-html="true"></div>
                            </div>
                        </div>
```
but instead we should have components and write:
```html
<div sd-ui-column-box data-large="true" data-dark-ui="true" data-label="{{ :: 'Metadata | translate }}">
    <div sd-ui-slide-in-panel data-first="true">
         <div sd-ui-alert data-hollow="true" data-primary="true" data-small="true">
             <span translate>Currently editing: {{images.length}} items</span>
             <div sd-ui-form-row ng-if="validator.headline">
                 <div sd-ui-line-input data-dark="true" data-boxed="true"
                      data-required="validator.headline.required"
                      data-label="{{ :: 'Title' | translate }}"
                      data-type="text"
                      data-model="metadata.headline"
                      data-change="onChange('headline')"
                      data-blur="onBlur()"
                      data-disabled="!images.length"
                      data-placeholder="{{placeholder['headline']}}"
                      data-maxlength="validator.headline.maxlength"
                  ></div>
```
so we would be building the views from components rather then writing html. then we might have consistent ui and if there is some change in ui framework you just fix the components and not all the views that might be using it. there is so much boilerplate otherwise that it's hard to figure out what's going on.

petrjasek:
and then one step forward would be something like:
```html
<div sd-ui-form data-form="form"></div>
```
and
```js
scope.form = {
    rows: [
        {type: 'text', label: 'Title', model: 'metadata.headline', ...},
    ],
}
```

pavlovicnemanja:
I can't agree more with you... This will be ideal way to use UI Framework 🙂 I will just keep components in UI Framework, it's smaller and easier to organize... Also, saying again, now it's much more easier to create UI prototypes when you have all this on UI framework.

fritzSF:
Yes, that would be nice. I do see a benefit in having the components it the UI framework though - it makes prototyping of complex things, like Planning and Photo desk, really easy to do.

petrjasek:
fair enough, as long as there are components I don't care where it is :wink:

PR #252, opened 2018-09-25
Development build of react is being used warning is present on sd master (SDESK 3441)

tomaskikutis:
SDESK-3441

`superdesk-ui-framework` was always built in development mode. I've changed it to build for production by default and added a grunt task to build for development.


petrjasek:
I'm not sure if we should add dist files into the repo, in theory we can run the build on install or on publish

tomaskikutis:
Yeah, we shouldn't. I'll try to remove it and see if `postinstall` script can build it. `publish` might also work, but it currently takes manual work to release to npm. Until we have a build server watching GitHub releases and automatically creating the same ones on `npm`  I'd like direct installs from GitHub to work as well.

tomaskikutis:
I wasn't able to find a clean solution for running a build after install. It builds after install locally, but not when you run the install from another project which depends on this one. The closest viable solution I found was this https://github.com/lzrski/npm-git-install, but I think we would be better off simply releasing to npm for now and hopefully automating it to mirror GitHub releases in the future.

petrjasek:
I will check, it should be possible to do it via travis

pavlovicnemanja:
Also, we are using dist files for doc page, because currently it is implemented as github page. It will be great if we can avoid this also, so dist is there also build automatically. It will avoid conflicts on dist files.

tomaskikutis:
We need another bot to pull `dist` folder from npm releases and publish it as a GitHub page :laughing: I don't know what to do about that, I suggest simply checking documentation locally for now and bringing the GitHub page down. Any other ideas?

fritzSF:
@tomaskikutis I disagree that we don't need the documentation page. I don't really care how we do it but let's find a way to have it.

MarkLark86:
@tomaskikutis I agree with @fritzSF We use the github doc page for our reference guide to the latest version/functionality when using the UI Framework

tomaskikutis:
The page is restored. We'll remove the `dist` folder from the repo after we have the automation for publishing docs.

PR #253, opened 2018-09-26
setup travis for code style checking

petrjasek:


app/scripts/helpers/taginput.helper.js
```diff
@@ -1050,69 +1069,78 @@
...
+        utils.debounce = function(fn, delay) {
             var timeoutId;
+
             return function() {
                 var args = arguments;
+
                 $timeout.cancel(timeoutId);
-                timeoutId = $timeout(function() { fn.apply(null, args); }, delay);
+                timeoutId = $timeout(() => {
+                    fn(...args);
```
tomaskikutis: It's might not matter in this case, but a correct translation would be `fn.call(null, ...args)`. When you do `fn(...args)` you are not binding `this` and the function might receive `window` or `undefined` as  this when it used to get `null` previously.

petrjasek: this part was done by `eslint --fix` I hope they know what they are doing :wink:

PR #259, opened 2018-09-27
update npm version used on travis

petrjasek:


.travis.yml
```diff
@@ -3,6 +3,8 @@ node_js: '6'
 cache:
   directories:
   - "$HOME/.cache/yarn"
+before_install:
+- npm i -g npm@6
```
tomaskikutis: I hope it's smart enough not to reinstall npm if it's already at version 6. If it's not it might be useful to add a version check

PR #270, opened 2018-10-21
Fix editor popup scrollable areas

tomaskikutis:
SDESK-3359
REQUIRED FOR: https://github.com/superdesk/superdesk-client-core/pull/2606

### DIFF
--- a/app/styles/components/_sd-editor-popup.scss
+++ b/app/styles/components/_sd-editor-popup.scss
@@ -17,7 +17,6 @@ $editor-popup-padding: 1.6rem;
 
 .editor-popup__main {
     display: flex;
-    flex-direction: column;
     color: $sd-text;
     width: 32rem;
     max-height: 100%;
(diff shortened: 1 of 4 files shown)


tomaskikutis:
The way how flex properties were laid out didn't support multiple flex areas. Replies would scroll, but the initial comment wouldn't as you can see in SDESK-3359.

I removed flex properties and moved it to `FluidRows`(on client core for now) component where you can have multiple scrollable areas and they would adjust accordingly to the amount of content. [see here](https://user-images.githubusercontent.com/2076953/47353415-e5139500-d6c4-11e8-9d3e-e9d453778d98.gif)

Do you think it makes sense to keep the component in ui-framework? If so I'll move `FluidRows` here as well. If not I can move editor popup styles to the client.

fritzSF:
OK, thanks for the clarification Tomas.

PR #278, opened 2018-11-05
1.5.2

MarkLark86:


MarkLark86:
@fritzSF @pavlovicnemanja @petrjasek 
We require a bump in the ui-framework version so that my changes to the sd-tag-input are available in production/test instances.

Please let me know if I have done this wrong.
Many thanks

PR #296, opened 2018-11-28
Small CSS fixes

fritzSF:


app/styles/_helpers.scss
```diff
@@ -266,9 +266,56 @@ time {
...
-}
\ No newline at end of file
+}
+
+
+// Common margin overrides
+// --------------------------------------------------
+
+// top
+.sd-margin-t--1 {
```
tomaskikutis: What kind of design are you doing so you need such rules? :)

fritzSF: Overrides for certain elements, to avoid creating to much modifiers for each element.

tomaskikutis: Can you show an example of a design where you would put the override otherwise?

fritzSF: Usage should be restricted, I prefer modifiers, but for certain elements we end up needing too much options. This is for such cases, I would avoid using it to much

PR #350, opened 2019-05-20
Video editor UI

fritzSF:


app/styles/components/_sd-photo-preview.scss
```diff
@@ -81,6 +147,154 @@ $photo-nav-transition: all 0.2s ease-out;
     box-shadow: inset 0 2px 6px rgba(0, 0, 0, 0.2);
     position: relative;
     overflow-y: auto;
+    &--video {
```
tomaskikutis: I suggest to stop using `&` operator. It's impossible to find which styles are applied to a class name. In markup, you'd see `class="sd-photo-preview__image--video"`, but there's no such class in sass code.

PR #357, opened 2019-06-17
feat(2.0): Created infrastructure update for ui-framework, added react playground and react components

pavlovicnemanja:


examples/reactTemplates/LineInput.tsx
```diff
@@ -0,0 +1,117 @@
...
+    componentDidMount() {
+        Prism.highlightAll();
+    }
+
+    render() {
+        return (
+            <div id="line-input">
+                <h2 className="docs-page__h2">Line Input</h2>
+                <pre className="line-numbers">
+                    <code className="language-tsx">{`<LineInput> {InputContent} </LineInput>`}</code>
```
tomaskikutis: The example is incorrect. The component doesn't accept children.

tomaskikutis:
* React components don't have interfaces and can't be checked by TypeScript. IntelliSense won't able to work either.
* Build doesn't work due `import './style.scss';` in `app/components/Form/TextArea.tsx`
* build doesn't work with node <8.6. If I'm not mistaken, we are currently using node 6 in client-core.
* even when I commented out the style import and used node 9, the build still didn't work saying `Module parse failed: The keyword 'interface' is reserved (4:0)`. Looks like TypeScript files weren't compiled.
* There aren't any live examples of the components. It might be worth looking into https://storybook.js.org or a similar tool so it's easier to document the components and create interactive live demos.
* React components need to be cleaned up. The purpose of ui framework is to be used across multiple projects. If we don't need that, let's not put them in the framework since multiple projects using them will make them harder to refactor. If we are to put components there, they should have clean interfaces without stuff like `paddingRight60`, `nativeOnChange`, `autoHeightTimeout`.

```
withButton | boolean | false | Adds class `sd-line-input--with-button`
```
Documentation should be better. As a consumer of the framework, there's no benefit to me to know about the class names. It's an implementation detail that should stay in the framework. What needs to be described is what the prop does.

examples/reactTemplates/LineInput.tsx
pavlovicnemanja: It is accepting. `LineInput` is only wrapper. 
image

pavlovicnemanja:
- This is just beginning of creating UI React components, I'm preparing interfaces. First of first is to create and separate React components. It is little slower, because I'm still learning and getting used to React.
- Build fixed, forgot to remove style from file.
- Not sure why are we limited to certain version of nodejs? Build is working on my local instance, version v10.10.0.
- UI framework is built to separate stuff and make cleaner code on `superdesk-client-core`. It was created before `superdesk-publisher`, `superdesk-planning` and other addons. After that, idea here was that other standalone projects can use our styles and designs.
- UI framework is not created for everyone, it is created as helper for superdesk-client-core, and thing to make code cleaner and smaller. Also, main advantage of superdesk-ui-framework is playgrounds. Designers here can make completely working prototypes of new features. Also, developers can use same classes and components to build it in application. New idea with react components is also to create this playground prototypes with React components.

examples/reactTemplates/LineInput.tsx
tomaskikutis: Sorry, my bad. I think I wanted to comment on `<Label>Some label</Label>` example, which doesn't accept children.

tomaskikutis:
> This is just beginning of creating UI React components, I'm preparing interfaces. First of first is to create and separate React components. It is little slower, because I'm still learning and getting used to React.

I think it's better to stash the React components for now if they aren't ready. Doing things gradually is fine at the application level, but it doesn't work that well for libraries. When you release library code, multiple projects will start depending on it, and you won't be able to refactor it without breaking the code of others. Implementation can be improved gradually, but interfaces should be well defined because that's what consumers will depend on.

> Build fixed, forgot to remove style from file.
:+1: 

> Not sure why are we limited to certain version of nodejs? Build is working on my local instance, version v10.10.0.

It doesn't work on versions lower than 8.6. In theory, ui framework doesn't have to work on node 6, but it will be harder to work with it if we have to switch between versions when working on client-core and ui-framework. It might not be a problem if client-core works with newer versions too.

> idea here was that other standalone projects can use our styles and designs
> UI framework is not created for everyone, it is created as helper for superdesk-client-core

I'm lost :D So do we want to expose these React components for other projects or not? If we don't it's better to keep it in the core. When it's in the core, it's easier to maintain and refactor since you don't need to check if other projects depend on it and you don't need multiple PRs to solve a single problem. Cleaner code comes from hiding implementation details and creating boundaries in the code which is accomplished by a separate build process in case of UI framework. The code won't become less clean whether it is in one repository or another.

tomaskikutis:
The build still doesn't work. Here's what I did:
```
tomas@p50:~$ node -v
v9.11.2
cd superdesk-ui-framework
rm -rf node_modules
yarn install
npm run build

cd superdesk-client-core
rm -rf node_modules
yarn install
grunt server
```


```
ERROR in ../WEBPACK_IGNORE/superdesk-ui-framework/app/components/Form/Checkbox.tsx
Module parse failed: The keyword 'interface' is reserved (4:0)
You may need an appropriate loader to handle this file type.
| import classNames from 'classnames';
| 
| interface IProps {
|     field?: string;
|     label?: string;
 @ ../WEBPACK_IGNORE/superdesk-ui-framework/app/index.js 11:0-52
 @ ./scripts/vendor.ts
 @ ./scripts/index.ts
 @ multi (webpack)-dev-server/client?http://0.0.0.0:9000 ./scripts/index

ERROR in ../WEBPACK_IGNORE/superdesk-ui-framework/app/components/Form/Label.tsx
Module parse failed: Unexpected token (10:18)
You may need an appropriate loader to handle this file type.
|  * @description Form label component
|  */
| export const Label: React.StatelessComponent<any> = ({text, row, light, invalid}) => (
|     !text ? null : (
|         <label className={classNames({
 @ ../WEBPACK_IGNORE/superdesk-ui-framework/app/index.js 9:0-46
 @ ./scripts/vendor.ts
 @ ./scripts/index.ts
 @ multi (webpack)-dev-server/client?http://0.0.0.0:9000 ./scripts/index
```

When this is fixed, I'd like you to stash React components for now and only commit the infrastructure part in this PR. As I mentioned, the components need good interfaces so let's wait until we have them so people don't depend on code that we will likely soon change.

pavlovicnemanja:
@tomaskikutis as you can check in package.json, we don't need `npm run build` anymore, because entry file is app/index.js. So we can import separate components in superdesk-client-core. `npm run build` and `dist` folder is now used only for `netlify` and displaying docs content on https://ui-framework.superdesk.org .

I don't want to stash react components because we need them so we can start creating useful design examples and playgrounds, as we are doing currently with angular components. As you can see, we have completely replicated playgrounds were we create and design new components. No one is using superdesk-ui-framework except us for now, so there is no fear that someone is using this components in wrong way. @fritzSF and @darconny will also work on design improvements of certain components.

We can't build everything right away. But this is very beginning of migrating superdesk-ui-framework from `angularjs` to `react`.

tomaskikutis:
The errors I've pasted aren't because of CSS, but because TypeScript files in ui-framework aren't compiled and webpack on client-core is configured not to compile anything inside `node_modules`. I think the most convenient way to solve this, would be to add a post-install script in ui-framework which would compile the TypeScript files.

tomaskikutis:
There are still a few issues:

* Can't import CSS from client-core

```
ERROR in ../WEBPACK_IGNORE/superdesk-ui-framework/dist/superdesk-ui.bundle.js
Module build failed: Error: ENOENT: no such file or directory, open '/home/tomas/DevTools/superdesk-repos/WEBPACK_IGNORE/superdesk-ui-framework/dist/superdesk-ui.bundle.js'
```

* Can't import `HelloWorld` component from client-core (I've added a commit to your branch)
* There are TypeScript errors in `HelloWorld` component and build still passes.

PR #365, opened 2019-07-16
Added canChangeStep method to wizard. (SDESK-4325)

GyanP:
canChangeStep method will prevent from step change when required.

### DIFF
--- a/app/scripts/wizard.js
+++ b/app/scripts/wizard.js
@@ -71,15 +71,23 @@ function WizardDirective() {
                 }
 
                 $scope.goTo = function (step) {
-                    $scope.nextStep = step.code;
-
-                    if ($scope.canChangeStep()) {
+                    const execute = () => {
                         unselectAll();
                         $scope.selectedStep = step;
                         if (!_.isUndefined($scope.currentStep)) {
                             $scope.currentStep = step.code;
                         }
                         step.selected = true;
+                    };
+
+                    if (typeof $scope.canTabChange === 'function') {
+                        $scope.canTabChange().then((result) => {
+                            if (result === true) {
+                                execute();
+                            }
+                        });
+                    } else {
+                        execute();
                     }
                 };
 
@@ -115,17 +123,6 @@ function WizardDirective() {
                         $scope.finish();
                     }
                 };
-
-                this.goToNextStep = function() {
-                    this.goTo($scope.nextStep);
-                }
-
-                $scope.canChangeStep = function () {
-                    if ($scope.canTabChange) {
-                        return $scope.canTabChange();
-                    }
-                    return true;
-                }
             }]
     };
 }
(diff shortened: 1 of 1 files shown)


app/scripts/wizard.js
```diff
@@ -70,12 +71,16 @@ function WizardDirective() {
                 }
 
                 $scope.goTo = function (step) {
-                    unselectAll();
-                    $scope.selectedStep = step;
-                    if (!_.isUndefined($scope.currentStep)) {
-                        $scope.currentStep = step.code;
+                    $scope.nextStep = step.code;
+
```
tomaskikutis: Why do you need `$scope.nextStep`, `this.goToNextStep`, `$scope.canChangeStep`. Doesn't it work to call `$scope.canTabChange` from `this.goTo` directly?

GyanP: `this.goTo` is called when we want to go to a certain step without clicking on the tab, otherwise `$scope.goTo` is called directly on clicking a tab. `$scope.nextStep` holds the next steps name, so when the user closes the confirmation prompt and wants to change the step, then `$scope.goToNextStep` method takes him to the step he clicked earlier.

GyanP: This is how `canTabChange` method looks
code_cut
 In this method, if in place of `$q.when(true)`, I try to use `Promise.resolve(true)` then I had to double click on the same tab to get it changed. I don't understand why it's happening. Do you have any idea?

tomaskikutis: Firstly, I would avoid using angular $q promises together with native promises. It might not be the issue, but it's a good practice not to mix things. Prefer native promises where possible.

It's also strange that when `confirmSave` is resolved, you are resolving `false` from `canTabChange`. If you still can't figure it out, push the non-working code and I'll take a look.

GyanP: Sorry, my bad, that `Promise.resolve(confirmSave)` is not there, by mistake I left it there. The issue is with `Promise.resolve(true)` instead of `$q.when(true)`.

tomaskikutis: I don't see any other obvious mistakes. If you can't figure it out, post the commit tags of ui-framework and client-core and I'll check it out.

GyanP: client-core: `eb6a559`
ui-framework: `51b1a53`

tomaskikutis: I don't see a PR in client-core. Can you push the code so I can check it out?

GyanP: sorry, this is the PR https://github.com/superdesk/superdesk-client-core/pull/3000

tomaskikutis: screenshot_1

I found the issue. You need to add `$scope.$apply();` to tell angular to re-render. You don't need that when using angular $q promises, because they are coupled with angular and always tell it to re-render when promises resolve.

PR #366, opened 2019-07-19
Basic size modifications

fritzSF:


app/styles/_helpers.scss
```diff
@@ -382,129 +397,116 @@ time {
...
-    margin-right: 2rem !important;
+// -------------------------------------------------
+
+// $sd-base-increment - default size of the increment is 0.8rem;
+// For example: .sd-margin-t--1 {margin-top: 0.8rem}, .sd-padding-l--2 {padding-left: 1.6rem}
+$sd-increments: ('0': 0, '0-5': 0.5, '1': 1, '1-5': 1.5, '2': 2, '3': 3, '4' : 4, '5' : 5); // Number of increments, genarates new classes eg. .sd-padding-top--4, .sd-margin-top--4
+$sd-sides: (top, right, bottom, left, x, y); // It's generating these sides
+$sd-properties: (padding, margin); // It's generating these properties
+
+@each $incrementName, $incrementValue in $sd-increments {
```
tomaskikutis: What benefits do these bring compared to exporting `sd-base-increment` to javascript and using inline styles like `<div style={{paddingLeft: sdBaseIncrement * 3}}></div>` ?

app/styles/_helpers.scss
```diff
@@ -382,129 +397,116 @@ time {
...
+        @each $side in $sd-sides {
+            @if($side == x) {
+                .sd-#{$property}-#{$side}--#{$incrementName} {
+                    #{$property}-left: #{$sd-base-increment * $incrementValue} !important;
+                    #{$property}-right: #{$sd-base-increment * $incrementValue} !important;
+                }
+            } @else if ($side == y) {
+                .sd-#{$property}-#{$side}--#{$incrementName} {
+                    #{$property}-top: #{$sd-base-increment * $incrementValue} !important;
+                    #{$property}-bottom: #{$sd-base-increment * $incrementValue} !important;
```
tomaskikutis: You shouldn't use `important` for base styles.

app/styles/_helpers.scss
fritzSF: Consistency, and building a design system based on the default increment.

app/styles/_helpers.scss
fritzSF: Those are not base styles but helpers that need to override all set styles. it's common practice in UI frameworks for helper classes. It has to stay as it is.

app/styles/_helpers.scss
tomaskikutis:
```
const getSpacing = (multiplier) => (sdBaseIncrement * multiplier) + 'rem';
a = <div style={{paddingLeft: getSpacing(3)}}></div>
```

You get the same consistency by exporting the increment, don't you?

What design system do you intend to build? Isn't UI framework a collection of independent components? They can all use the `sdBaseIncrement` variable without having to have a layer of generated single-property class names.

fritzSF: A design system for Superdesk – I'll gladly explain to you the general idea outside of the code review. The UI framework will be an integral part of the design system. We do need classes to address the proper usage directly in the design system. It will also be used trough CSS for other classes (eg. @include sd-margin('3', 'top')}.

PR #389, opened 2019-11-07
feat(react): added buttons, labels and badges components

gagily:


app-typescript/components/Button.tsx
```diff
@@ -0,0 +1,47 @@
+import * as React from 'react';
+import classNames from 'classnames';
+
+interface IProps {
```
tomaskikutis: Implement icon-only buttons and standard text buttons separately according to these interfaces:

```
interface IButtonBase {
    theme: 'light' | 'dark';
    icon?: 'success' | 'alert'; // <-- add available icon names
    disabled?: boolean; // defaults to false
    type?: 'default' | 'primary' | 'success' | 'warning' | 'alert' | 'highlight' | 'sd-green';
    size?: 'small' | 'normal' | 'large'; // defaults to normal
    tooltip?: string;
    children?: never;
}

interface IPropsButton extends IButtonBase {
    text: string;
    expand?: boolean; // defaults to false
    style?: 'filled' | 'hollow' | 'text-only'; // defaults to filled
}

interface IPropsIconButton extends IButtonBase {
    shape?: 'square' | 'round'; // defaults to square
}
```

This is for taking advantage of the compiler to only allow valid configuration options. For example, the interface in this file has many properties and it's not obvious which can't be used together like`textOnly=true iconBtn=true`. Having separate components makes it possible to validate component configurations it at compile time.

app-typescript/components/Label.tsx
```diff
@@ -0,0 +1,38 @@
+import * as React from 'react';
+import classNames from 'classnames';
+
+interface IProps {
```
tomaskikutis: Can you use the following interface?

```
interface IProps {
    text: string;
    icon?: 'icon1' | 'icon2'; // <-- add available icons

    // I suggest not exposing a prop for text transform and handling it inside the component
    // small and large would be uppercase and medium - sentence case
    size?: 'small' | 'normal' | 'large';
    color: 'color1' | 'color2'; // <-- add available colors
    style?: 'filled' | 'hollow'; // defaults to filled
    onClick?(): void; // if onClick is provided, a "link" icon is displayed
}
```

Does it make sense to have a icon-only label? If it does, add a separate component like with icon-only buttons. I wasn't able to understand what interface property `types` is referring to. Can you explain?

app-typescript/components/TagLabels.tsx
```diff
@@ -0,0 +1,26 @@
+import * as React from 'react';
+import classNames from 'classnames';
+
+interface IProps {
```
tomaskikutis: Can you use the following interface?

```
interface IProps {
    text: string;
    shape?: 'square' | 'round'; // defaults to round
    color: 'gray' | 'blue';

    // makes color more saturated
    highlighted?: boolean; // defaults to false

    // adds x button for removing the label
    onClick?(): void;
}
```

app/index.js
```diff
@@ -7,7 +7,13 @@ export {Positioner} from './scripts/positioner';
 
 // TypeScript exports
 // Must contain the same exports as in '../app-typescript/index'
-export {HelloWorld} from '../app-typescript/index';
+export {
+    HelloWorld,
+    Button,
+    Label,
+    Badge,
+    Alert,
```
tomaskikutis: Add TagLabel

gagily: In agreement with @fritzSF, we are removing that component

tomaskikutis:
I see you have forced pushed come commits. It's not an issue yet because you did it before I started reviewing the code, but please don't rebase/force push anymore so git history is preserved and I can review your upcoming commits individually.

app-typescript/components/Label.tsx
gagily: it's impossible to define all colors and icons  because there is a lot of them and there are being constantly updated

tomaskikutis: make it a `string` then, and add a comment with a link to a file where all defined values are listed.

gagily: here is a link to icons https://ui-framework.superdesk.org/#/components/icons, and to colors https://ui-framework.superdesk.org/#/components/colors

tomaskikutis: Add the links to the interface like so

```
interface IProps {
    // https://ui-framework.superdesk.org/#/components/icons
    icon?: string;
    ...
}

I'm not sure the colors in that link are implemented. If they are, specify the exact format for a color string. For example, is it "PURPLE500" or "purple_500"?

File: app-typescript/components/Label.tsx [hunk 5]
```diff
@@ -3,31 +3,31 @@ import classNames from 'classnames';
...
+    size?: 'small' | 'normal' | 'large';
+    onClick?(): void;
     noTransform?: boolean;
-    types?: string;
-    typesLabelIcon?: string;
-    hollow?: boolean;
+    style?: 'filled' | 'hollow';
     iconLabel?: boolean;
-    icon?: string;
+    typesLabelIcon?: string;
```
tomaskikutis: Why do you need 3 icons? `icon`, `iconLabel` and `typesLabelIcon`?

app-typescript/components/Alert.tsx
```diff
@@ -0,0 +1,54 @@
...
+        super(props);
+        this.state = {
+            open: true,
+        };
+        this.onClose = this.onClose.bind(this);
+    }
+
+    onClose() {
+        this.setState((state) => ({
+            open: !state.open,
```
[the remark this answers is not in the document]
pavlovicnemanja: Leave icon as icon, because initial state of alert can also be closed, so restore-icon doesn't have sense. Toggle is not enough for knowing that alert can be opened and closed?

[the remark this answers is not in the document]
tomaskikutis: > Toggle is not enough for knowing that alert can be opened and closed?

`onToggle` would be a part of the implementation. When you are trying to use a component from client-core and jump to type definitions, you don't see the implementation - only types. And that's a good thing. Types and comments on types should be enough to understand how to use the component.

I'd prefer to call it `restore-icon` because it's only displayed when the alert is hidden which is not obvious otherwise. For example, when I was initially testing the component, I didn't know it can be restored after closing and was thinking that the icon doesn't work - because I expected the icon to be visible somewhere next to the alert text.

fritzSF:
> You also need to fix component isolation as discussed [here](https://sourcefabric.slack.com/archives/C3CU0BUFQ/p1570529314000300). Theming is not important at the moment since it can be easily changed later without affecting anything else, but the spacing and class name aliases need to fixed now. You don't need to namespace class names(second point) which are not overwritten in client-core stylesheets.

If you are referring to paddings and margins in the alert they should stay. There should be no option to change this. I would treat this the basically the same as buttons. The bottom margin is basically mandatory for all use cases. Anyway, I don't see any of this as a blocker.

tomaskikutis:
I'm not talking about alerts specifically. I'm referring to all components that have spacing. The reason I want this is for maintenance to be easier. The inside of the component and spacing around it are issues that can be solved separately. Coupling them together doesn't bring any significant benefits, but makes refactoring much harder.

If it's easier that way, you can keep the spacing in the framework and I'll wrap the components in the client and remove the spacing myself.

Namespacing class names is a must though since we won't be able to use the components in some parts of the Superdesk where these class names are overwritten.

fritzSF:
It is not about being easier, it's about consistency. If a component is designed to have always the same padding value it will stay in the framework and you or anybody else should not make any overrides in the client. That's one of the main reason we are doing this in the first place.  If there are reasons to modify such values in a component it can and should be added as a property or handled in a different way in the framework (e.g. button-group for buttons).

fritzSF:
> Namespacing class names is a must though since we won't be able to use the components in some parts of the Superdesk where these class names are overwritten.

It's not a must in all cases, only for components that are tight already deep into core.

tomaskikutis:
Consistency is easily achieved with spacer components or button groups and we don't need to add coupling in order to achieve it.

> It's not a must in all cases, only for components that are tight already deep into core.

Who's checking the tightness? :smile: Let's not have any overwrites in the client-core. If it's not "deep in the core", remove the overwrites from the client. If it is - namespace it from the UI framework.

app-typescript/components/Label.tsx
```diff
@@ -3,31 +3,31 @@ import classNames from 'classnames';
 
 interface IProps {
     text?: string;
-    link?: string;
-    sizes?: string;
+    icon?: string; // https://ui-framework.superdesk.org/#/components/icons
+    color?: string;  //  https://ui-framework.superdesk.org/#/components/colors
+    size?: 'small' | 'normal' | 'large';
+    onClick?(): void;
     noTransform?: boolean;
```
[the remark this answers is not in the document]
tomaskikutis: This is no longer relevant. We need an option to be able to display large buttons either as uppercase or lower case.

app-typescript/components/Label.tsx
```diff
@@ -3,19 +3,19 @@ import classNames from 'classnames';
 
 interface IProps {
     text?: string;
-    color?: string;  //  https://ui-framework.superdesk.org/#/components/colors
+    color?: string;
     size?: 'small' | 'normal' | 'large'; // defaults to 'normal'
     onClick?(): void;
-    noTransform?: boolean;
+    transform?: boolean;
```
tomaskikutis: Can you restore the old name (`noTransform`)? The new one isn't accurate, because you would pass `transform={true}` and it wouldn't get transformed.

app-typescript/components/IconButton.tsx
```diff
@@ -0,0 +1,21 @@
+import * as React from 'react';
+
+interface IProps {
+    tooltip?: string;
```
tomaskikutis: Use the same tooltip format as in IconButton

```
tooltip?: {
    text: string;
    flow?: 'top' | 'left' | 'right' | 'down'; // defaults to 'top'
};
```

app-typescript/components/NavButton.tsx
```diff
@@ -0,0 +1,24 @@
+import * as React from 'react';
+import classNames from 'classnames';
+
+interface IProps {
+    icon?: string;
+    theme?: 'light' | 'dark'; // defaults to 'light'
+    color?: string;
```
tomaskikutis: add `disabled?: boolean;` and `onClick(): void;` properties.

app-typescript/components/Button.tsx
tomaskikutis: When I was referring to "icon only buttons" in the previous comment I meant "Buttons with icon font only"(btn--icon-only). Keep current `IconButton` as is, but create a separate component which would be for "Buttons with icon font only" and would use `IPropsIconButton` interface.

app-typescript/components/Tooltip.tsx
```diff
@@ -0,0 +1,21 @@
+import * as React from 'react';
+
+interface IProps {
+    text?: string;
+    tooltip?: string;
+    flow?: 'top' | 'left' | 'right' | 'down'; // defaults to 'top'
+}
+
+export class Tooltip extends React.Component<IProps> {
```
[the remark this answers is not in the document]
gagily: There was solution like that but it's decided in favor of good performance to be removed. Discuss with @pavlovicnemanja and @fritzSF .

app-typescript/components/Button.tsx
```diff
@@ -2,25 +2,22 @@ import * as React from 'react';
...
 interface IPropsButton extends IButtonBase {
     text?: string;
     expand?: boolean;
-    style?: 'filled' | 'hollow' | 'text-only';
-}
-
-interface IPropsIconButton extends IButtonBase {
-    shape?: 'square' | 'round';
+    style?: 'filled' | 'hollow' | 'text-only'; // defaults to 'filled'
+    shape?: 'square' | 'round'; // defaults to 'square'
```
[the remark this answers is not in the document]
gagily: Button with icon only in button component can be round or squared so we need shape props.

app-typescript/components/IconLabel.tsx
```diff
@@ -0,0 +1,25 @@
+import * as React from 'react';
+import classNames from 'classnames';
+
+interface IProps {
+    text?: string;
+    color?: string;  //  https://ui-framework.superdesk.org/#/components/colors
```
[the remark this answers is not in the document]
gagily: There isn't file with listed available colors. Colors are defined in sass and created with function.

app-typescript/components/Button.tsx
gagily: In agreement with @pavlovicnemanja it's decides not to make separate component because it's going to be to much small components with only one props difference between them.

app-typescript/components/Tooltip.tsx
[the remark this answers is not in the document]
tomaskikutis: Only a single tooltip can be displayed at once, so it shouldn't affect performance. Anyway, let's keep it as is for now. I'm looking at adding components from other open source libraries so we'll probably be able to swap this component with the one that works with `overflow:hidden`.

app-typescript/components/IconLabel.tsx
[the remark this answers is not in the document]
tomaskikutis: Can you link to the sass file where available color strings are listed?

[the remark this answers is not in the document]
pavlovicnemanja: It is better to have link to docs page, it is more readable and feature developers will know which color to pick. If we add link to sass file it will mean nothing. Because colors are created with multiple sass functions and they will be changed in future. @gagily leave it as it is.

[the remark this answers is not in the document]
tomaskikutis: Docs page works as well. Can you then add a link to the docs page?

app-typescript/components/Badge.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+export class Badge extends React.Component<IProps> {
+    render() {
+        let classes = classNames('badge', {
+            [`badge--${this.props.type}`]: this.props.type,
+            'badge--square': this.props.style === 'square',
+        });
+
+        return (
+            <React.Fragment>
```
tomaskikutis: you don't need a fragment here:

```
if (this.props.children) {
    return (
        <div className='element-with-badge'>
            {this.props.children}
            <span className={classes}>{this.props.text}</span>
        </div>
    );
} else {
    return <span className={classes}>{this.props.text}</span>};
}
```

app-typescript/components/Label.tsx
```diff
@@ -0,0 +1,32 @@
...
+    render() {
+        let classes = classNames('label', {
+            [`label--${this.props.size}`]: this.props.size !== 'normal' && this.props.size !== undefined,
+            'label--no-transform': this.props.noTransform,
+            [`label--${this.props.color}`]: this.props.color !== undefined,
+            'label--hollow': this.props.style === 'hollow',
+        });
+
+        return (
+            <React.Fragment>
```
tomaskikutis: You don't need a fragment here:

```
if (this.props.onClick) {
    return (
        <a href='' className={classes}>
            {this.props.text}
        </a>
    );
} else {
    return (
        <span className={classes}>
            {this.props.text}
        </span>
    );
}
```

app-typescript/components/IconLabel.tsx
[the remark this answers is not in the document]
tomaskikutis: Not that docs page :smile: This is the same you linked before which I commented not to link because Vladimir said it's for internal use only. Also, it doesn't represent the actual values. Here's the example code from the docs you've added and none of the colors you use are specified in that docs page.

```
<IconLabel text='Label primary'  icon='facebook' color='primary'/>
<IconLabel text='Label success' icon='text' color='success'/>
<IconLabel text='Label warning' icon='paywall' color='warning'/>
<IconLabel text='Label alert' icon='trash' color='alert'/>
<IconLabel text='Label highlight' icon='composite' color='highlight'/>
<IconLabel text='Label sd-green' icon='video' color='sd-green'/>
```

app-typescript/components/Button.tsx
tomaskikutis: You can reuse the same internal component. See this comment https://github.com/superdesk/superdesk-ui-framework/pull/389/files#r347453133

app-typescript/components/IconLabel.tsx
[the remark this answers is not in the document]
fritzSF: @tomaskikutis I think you got this wrong. It is completely fine as it is now. I don't know what internal use means in this case, the whole framework is for internal use. 'primary', 'success' etc. are semantic colours with appropriate names, colours listed in the extended colour library don't have any semantic meaning but are in this case needed for differentiation between different labels . As you know labels are in our case very specific and as we use them to represent different things with it.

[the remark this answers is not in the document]
tomaskikutis: As we can see in the example snippet, the following strings are available as color a parameter for `IconLabel` component - `primary, success, warning, alert, highlight, sd-green`.

My question is, what are all possible color names? Is it only these 6? If so, I want the component interface to say so like `color: 'primary' | 'success' | 'warning' | 'alert' | 'highlight' | 'sd-green'''. Otherwise, I'd expect a link with available color names listed.

app-typescript/components/Button.tsx
pavlovicnemanja: Why? It is easier just to define `shape`... We don't want milion small components in UI-Framework. We want easy to use components. Let's keep it as this for now. Buttons are buttons.

app-typescript/components/IconLabel.tsx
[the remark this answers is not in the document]
pavlovicnemanja: Primary, Success, Warning, Alert are default colors (similar as bootstrap), and also, if you apply `red-500` it will work.

app-typescript/components/Button.tsx
tomaskikutis: > Why?

Because `shape` doesn't work with text buttons by design. It's easier to understand and use the components when the interfaces are clean and all options work together. That's the main advantage against trying to put the components together by using CSS classes which might or might not work in different combinations. There might be cases when it's not possible to validate it with types only, but it's possible and quite easy in this case.

screenshot_14

What's the issue with multiple components? It won't be a million, only a few more. Some of them won't even be exported like `ButtonInternal` example in the comment I linked.

Components will be used significantly more often than they will be modified, so it makes sense to prioritize ease of use for end-users(people using it in client-core or publisher) rather than developers of the framework who are already more familiar with how it works internally.

app-typescript/components/Button.tsx
```diff
@@ -2,6 +2,7 @@ import * as React from 'react';
 import classNames from 'classnames';
 
 interface IButtonBase {
+    navClass?: string;
```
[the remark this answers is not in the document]
tomaskikutis: > It's easier to create two new components than just to define some prop in interface?

As you can see, it's 20 lines of code per component, so both options are easy. I assume you'd agree that exporting properties that shouldn't be used or don't work isn't a good design. Why not do it properly if both approaches are easy to do?

fritzSF:
This is taking way too long. To put this extremely long and exhausting discussion to an end I will make a decision that we won't use a separate component for icon-only buttons. This is a matter of approach and not code quality. I don't see this as a blocker and I want to see us moving forward with this.

app-typescript/components/IconLabel.tsx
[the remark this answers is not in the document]
tomaskikutis: Have you tried it? It doesn't work for me. I use the following code, but the label is gray

```
<IconLabel text='Label' icon='video'  color='red-500'/ >
```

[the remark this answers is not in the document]
gagily: I've fixed it now. Also format of colors are 'red--500' with two dashes.

PR #390, opened 2019-11-20
feat(switch): Added switch component in react

pavlovicnemanja:


app-typescript/components/Switch.tsx
```diff
@@ -0,0 +1,48 @@
+import * as React from 'react';
+import classNames from 'classnames';
+
+interface IProps {
+    value?: boolean;
```
tomaskikutis: make value required

app-typescript/components/Switch.tsx
```diff
@@ -0,0 +1,48 @@
...
+}
+
+interface IState {
+    checked: boolean;
+}
+
+export class Switch extends React.Component<IProps, IState> {
+    constructor(props: IProps) {
+        super(props);
+        this.state = {
```
tomaskikutis: Remove the internal state and use `this.props.value` instead.

app-typescript/components/Switch.tsx
pavlovicnemanja: Why? If it is required you will need to specify value all the time. If it is not required, you can just go with default behaviour (false). It's cleaner if we go with natural approach.

app-typescript/components/Switch.tsx
pavlovicnemanja: If we keep current behaviour for `value` not to be required, then we will need to keep state.

app-typescript/components/Switch.tsx
pavlovicnemanja: But yea, if we specify model inside value, then it will automatically be changed when we change props value. Changing...

pavlovicnemanja: Then we can make onChange not required in this case

tomaskikutis: When you use the component, you'll have to provide `onChange` function, which will update the state in another component. It's already in the state elsewhere, so it's not useful having it twice.

When would you have `onChange` optional? What's the use of the switch if you can't change the value?

pavlovicnemanja: But `this.props` are readonly. How to change internal behaviour (adding class) if I don't have internal state? I will need to have internal state so I can set checked true or false and add corresponding class. Why I need to set `value` and `onChange` both required? `value` can go with default behaviour of false, and onChange will update state in another component.

pavlovicnemanja: I think I get what you want to say. This will be handled outside, `this.props.value` will be changed outside of this component

tomaskikutis: That's right. You will always have the newest value in `this.props`. See https://reactjs.org/docs/forms.html#controlled-components

app-typescript/components/Switch.tsx
```diff
@@ -26,16 +22,12 @@ export class Switch extends React.Component<IProps, IState> {
             return;
         }
 
-        this.setState((state) => ({
-            checked: !state.checked,
-        }));
-
-        return this.props.onChange(this.state.checked);
+        return this.props.onChange(!this.props.value);
```
tomaskikutis: You don't need a `return` here.

PR #399, opened 2019-11-28
feat(react): added radio and checkbox component

gagily:


app-typescript/components/Radio.tsx
```diff
@@ -0,0 +1,47 @@
+import * as React from 'react';
+
+interface IValue {
+    value: any;
```
tomaskikutis: Can you make it `string` instead of `any?` Try not using `any` at all. It's mainly meant for migrations or integration of libraries without typings. You shouldn't need it in your own code. You can always ping me if you get a compiler error you don't understand :slightly_smiling_face:

app-typescript/components/Radio.tsx
```diff
@@ -0,0 +1,47 @@
...
+    text: string;
```
tomaskikutis: `label` would be a more specific name

app-typescript/components/Radio.tsx
```diff
@@ -0,0 +1,47 @@
...
+    disabled?: boolean;
+}
+
+interface IProps {
+    model: any;
+    values: Array<IValue>;
+    side?: string;
```
tomaskikutis: `labelSide` would be a more specific name. Which sides are supported? Use union types:
```labelSide: 'side1' | 'side2';```

File: app-typescript/components/Radio.tsx [hunk 4]
```diff
@@ -0,0 +1,47 @@
...
+    onClick(nextValue: string): void;
```
tomaskikutis: Use `onChange` which is a more standard name. Most inputs can be changed via keyboard too.

app-typescript/components/Radio.tsx
```diff
@@ -0,0 +1,47 @@
+import * as React from 'react';
+
+interface IValue {
+    value: any;
+    text: string;
+    disabled?: boolean;
+}
+
+interface IProps {
+    model: any;
```
tomaskikutis: Can you use a more standard naming?

```
options: Array<IOption>;
value: string;
```

app-typescript/components/Radio.tsx
gagily: Value can be string, object or boolean, because of that I used any.

tomaskikutis: It doesn't make sense for value to boolean, because it wouldn't work to have more than 2 options. If you have 3 and their values are `true, false, true` and the current value is `true`, how would you know which label to mark as active?

Allowing objects isn't that good either because they can be compared by value or reference which might cause confusion when using the component. Can you limit allowed `value` type to `string`?

tomaskikutis:
Can you take a look at the unresolved comments? There are a few points you didn't address including the type of checkbox value property, checkbox `active` property, making checkbox side non-optional.

app-typescript/components/Radio.tsx
pavlovicnemanja: But sometimes, especially in settings we need to allow assigning object as value. Why is that big deal not to allow this if we need that? Also, `PrimeReact` that you used for `DatePicker` is allowing objects as values.

https://github.com/primefaces/primereact/blob/master/src/components/checkbox/Checkbox.d.ts

pavlovicnemanja:
@tomaskikutis why is important making checkbox side non optional? We shouldn't force people to enter side, by default it should be left, and especially in some certain cases if you want to change side you can specify it.

app-typescript/components/Radio.tsx
tomaskikutis: Can you show me an example of how would you use object values?

It doesn't matter what `PrimeReact` uses. I used the library because it was the best I could find at displaying the component. I could care less if they have bad interfaces because I wrap their components and don't expose their interfaces to consumers of `superdesk-ui-framework`.

pavlovicnemanja: If you want to assign some specific object to value, you will use object as value. Not sure what is not clear about that? It is matter, because here we are talking about what should be the best and what is best implementation. So let's check how big teams focused only on developing components did it.

tomaskikutis: > If you want to assign some specific object to value, you will use object as value. Not sure what is not clear about that?

Can you paste a simple code snippet? There might be multiple ways to implement it and I don't know what's on your mind.

pavlovicnemanja:
```
let object1 = {'prop1: 'vaue1', 'prop2': 'value2'};
let object2 = {'prop1: 'vaue1', 'prop2': 'value2'};
let object3 = {'prop1: 'vaue1', 'prop2': 'value2'};

<Radio value={this.state.value1} options={[{value:object1, text:'Default label'}, {value:object2, text:'Default label'}, {value:object3, text:'Default label'}} onClick={(value) => this.setState(() => ({ value1: value }))}/>
```

tomaskikutis:
> @tomaskikutis why is important making checkbox side non optional? We shouldn't force people to enter side, by default it should be left, and especially in some certain cases if you want to change side you can specify it.

You may be right. Let's leave it optional.

app-typescript/components/Radio.tsx
tomaskikutis: This is dummy code which doesn't make it clear how would you use it in the real world. Even in this case, it's easy to pass `object1.prop1` as a value which is string.

The problem with objects is comparison. How do you compare if objects are the same in order to determine which label should be marked as active? If you do `===` comparison, you are checking by reference and will get bugs when reference changes but values stay the same. You can try serializing the object to a string, but not all objects are serializable. By using strings as values we would be avoiding the issue entirely.

app-typescript/components/Checkbox.tsx
```diff
@@ -0,0 +1,46 @@
+import * as React from 'react';
+
+interface IProps {
+    value: any;
+    label: {
+        text: string,
+        side?: 'left' | 'right',
+    };
+    disabled?: boolean;
+    side?: string;
```
tomaskikutis: You've probably forgotten to remove this one after moving it to label

app-typescript/components/Checkbox.tsx
```diff
@@ -0,0 +1,46 @@
+import * as React from 'react';
+
+interface IProps {
+    value: any;
```
tomaskikutis: Why is value `any` instead of boolean?

app-typescript/components/Radio.tsx
pavlovicnemanja: Yes, and your value will be only vaue1 and you will miss entirely value2. 

You wanted example, I wrote you example. Previously we needed this functionality, to get object assigned as value on `sd-check`, and I'm sure we will need this in feature. Because of that I'm highly recommending to allow this options. This is not bug, if you want you can go with string, but also you will have ability to use something else (for example object).

This is feature not a bug, not sure why we are taking energy on something that is so stupid. As I sad, string is recommended, but also we need to make ability to allow objects as values.

app-typescript/components/Checkbox.tsx
pavlovicnemanja: Checkbox values should not be only `true` or `false`. Sometimes you need to assign certain value to checkbox. This is also supported by native checkbox: https://www.w3schools.com/tags/att_input_type_checkbox.asp

pavlovicnemanja: Also here we can have same discussion should this be object, boolean or string. It can be everything.

tomaskikutis: The checkbox will be either checked or not, that's why I suggest having a boolean. What I assume you would do if an object is passed is check if it's `null` and convert it to boolean that way. I don't see a reason to put it inside component logic since it's trivial to do from outside of the component:
```
<Checkbox value={myObject != null} onChange={...} />
```

pavlovicnemanja: No, sometimes on `value` you want to assign `test1` not just `true`. Check link I send it to you, as you can see, basic functionality

tomaskikutis: If you pass `'test1'`, how will you decide if the checkbox is supposed to be checked or not? By converting it to boolean while checking if it's a string and not empty? The same can easily be done outside the component identically to the object example:

```
<Checkbox value={!isEmtpy(str)} onChange={...} />
```

You are also losing type safety by allowing objects. What if an object is empty? What if it's an array, Set, Map? It's just easier to have the consumer of the component handle it and pass the boolean value.

pavlovicnemanja: What you will get as response on onChange? `True` or `False`, not actual value. As I sad, check what is native behaviour of checkboxes.

pavlovicnemanja: If I need object, I will use object. Check other frameworks, all are allowing objects and setting custom values for them. What are we? Smartest people on planet and we want to make life of developers difficult? Idea of `superdesk-ui-framework` is to make use of components easier, not to make it to complicate.

tomaskikutis: If you pass a boolean in as a value, you get a boolean out on `onChange` too.

Let's save time on reviews: if you can, you review the components yourself. If you want me to take a look, add me as a reviewer. I'll post suggestions and you take away from there and decide which suggestions you want to implement :wink:

pavlovicnemanja: Yes, and you want to have string as output of `onChange` so you can assign it to value. Simple as that. No, here we are discussing about stupid things, trivial, something that is already implemented as native functionality of same component. Let's not lose time on stupid discussions and trivial things. Suggestions and reviews are important, but not like this. Let's focus on functionalities on specific component, not on types and naming (I'm not saying it's not important, but it is not big deal if something is called `value` or `model`, also agreeing to have something consistant). But discussions like this are exhausting and it's just case how developer is seeing something. And at the end you are trying to find arguments just to justify your arguments. We are not kids, we have kids.

app-typescript/components/Checkbox.tsx
```diff
@@ -0,0 +1,45 @@
...
+    label: {
+        text: string,
+        side?: 'left' | 'right',
+    };
+    disabled?: boolean;
+    onChange(nextValue: boolean): void;
+}
+
+export class Checkbox extends React.Component<IProps> {
```
petrjasek: I think we should be using PureComponents as much as possible

app-typescript/components/Checkbox.tsx
petrjasek: I don't see why we would need a value property here. what about:
```
interface IProps {
    checked: boolean;
    onChange: (checked: boolean) => void;
}

Comment (tomaskikutis) [reply app-typescript/components/Checkbox.tsx]: That's the same as what I was suggesting, except that I was calling the `on/off` state `value: boolean` instead of `checked: boolean`. I don't mind which name is used.

File: app-typescript/components/Radio.tsx [hunk 9]
```diff
@@ -0,0 +1,47 @@
...
+        this.props.onChange(value);
+    }
+
+    render() {
+
+        return (
+            this.props.options.map((item: any, index: number) => (
+                <span className='sd-check__wrapper' key={index}>
+                    {this.props.labelSide === 'left' ?
+                        (item.disabled ? <label className='sd-label--disabled'>{item.label}</label> :
```
petrjasek: imo label should be a component too to avoid code duplication

app-typescript/components/Radio.tsx
petrjasek: I think even if we would need to support objects it would be good to list those options there and not use any,
because that not doing any check. undefined will work too so it won't catch typos.
so define value as `string  | number | {}`

app-typescript/components/Checkbox.tsx
pavlovicnemanja: Ok, so not handling any value inside `Checkbox` component, just visually adding checkbox sign. This is ok for `Checkbox` but for `Radio` we will need to support value. Because of consistency I tough to implement same thing in `Checkboxes` because they are pretty similar components. So for the end user not to be confused about implementation. One component is returning boolean and other one is returning value. To make it similar as much as possible.

app-typescript/components/Radio.tsx
petrjasek: btw I think we had some usecase with elasticfilters, something like:
```
filterOptions = [
    {label: 'Today', value: {range: {versioncreated: {gte: 'now/d'}}}},
    {label: 'This week', value: {range: {versioncreated: {gte: 'now-1w/d'}}}},
    {label: 'Last week', value: {range: {versioncreated: {gte: 'now-2w/d', lte: 'now-1w/d'}}}},
    ...
];
```

app-typescript/components/Checkbox.tsx
petrjasek: I like things consistent, but explicit is even better :wink:

pavlovicnemanja: image

pavlovicnemanja: https://gph.is/10qGOQg

app-typescript/components/Radio.tsx
tomaskikutis: Native select inputs only allow strings as values. Here's how it would look with elastic filters:

```
return (
    <select value={this.state.value}>
        {
            filterOptions.map((option) =>
                <option
                    value={option.label}
                >
                    {option.label}
                </option>)
        }
    </select>
);
```

It requires you to have an ID for a value and then you can get the object by that ID.

If you want to go with objects, I suggest using generics, because `{test: 1}` can't be assigned to `string | number | {}`, since it's not an empty object. Here's an interface with generics:

```
interface IRadioProps<T> {
    options: Array<T>;
    value: T;
    onChange(value: T): void;
    getLabel(item: T): string;
    getId(item: T): string;
}
```

I'd look similar to how `select2` [is implemented](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/core/ui/components/select2.tsx#L8)

petrjasek: right, it could be `string | number | object`. we could use generic too I would keep it simple:
```ts
interface IProps<T> {
    options: Array<{
        label: string;
        value: T;
    };
    value: T;
    onChange: (value: T) => void;
}
```

tomaskikutis: How would you check if values are equal? Compare labels? If we force labels to be unique, which makes sense for radios, we could drop `getId` method in my suggested interface and I'd be as simple. I'd prefer passing a lambda to get a label instead of mapping the array, but both are fine.

petrjasek: I would compare values by reference or via [lodash.isEqual](https://lodash.com/docs/4.17.15#isEqual)

app-typescript/components/CheckboxButton.tsx
```diff
@@ -0,0 +1,38 @@
...
+}
+
+export class CheckboxButton extends React.PureComponent<IProps> {
+    constructor(props: IProps) {
+        super(props);
+        this.handleChange = this.handleChange.bind(this);
+    }
+
+    handleChange() {
+        this.props.onChange(!this.props.checked);
```
petrjasek: I think this should only call this if not disabled

examples/pages/playgrounds/react-playgrounds/components/SubNav.tsx
```diff
@@ -0,0 +1,21 @@
+import * as React from 'react';
+import classNames from 'classnames';
+
+interface IProps {
+    color?: 'lighter' | 'darker';
+    children?: React.ReactNode;
+}
+
+export class SubNav extends React.Component<IProps> {
```
petrjasek: PureComponent..

examples/pages/playgrounds/react-playgrounds/components/SubNav.tsx
```diff
@@ -0,0 +1,21 @@
...
+    render() {
+        let classes = classNames({
```
petrjasek: seems like it's not used

app-typescript/components/CheckButtonGroup.tsx
```diff
@@ -0,0 +1,26 @@
+import * as React from 'react';
+import classNames from 'classnames';
+
+interface IProps {
+    orientation?: 'horizontal' | 'vertical'; // defaults to 'horizontal'
+    grid?: boolean;
+    align?: 'left' | 'right' | 'center'; // defaults to 'left'
+    children: React.ReactNode;
```
tomaskikutis: You don't need to put children in the interface unless you want to limit it. For example, if the component doesn't support `children` like a Button, you can specify `children?: never;` so the code doesn't compile if it's passed. Or you can make children accept a function like `children: () => number;`. If any React component works as children, you don't need to specify it in the interface.

examples/pages/playgrounds/react-playgrounds/components/SubNav.tsx
```diff
@@ -0,0 +1,21 @@
+import * as React from 'react';
+import classNames from 'classnames';
+
+interface IProps {
+    color?: 'lighter' | 'darker';
```
tomaskikutis: That should probably be `theme`

examples/pages/playgrounds/react-playgrounds/components/SubNav.tsx
```diff
@@ -2,14 +2,14 @@ import * as React from 'react';
 import classNames from 'classnames';
 
 interface IProps {
-    color?: 'lighter' | 'darker';
+    theme?: 'lighter' | 'darker';
```
tomaskikutis: All components should use the same names for themes. Until now all components use `theme?: 'light' | 'dark';`. Can you use the same ones and map it to class names inside the component?

app-typescript/components/RadioButton.tsx
```diff
@@ -0,0 +1,42 @@
...
+    value: any;
+    options: Array<IOption>;
+    labelSide?: 'left' | 'right';
+    onChange(nextValue: string): void;
+}
+
+export class RadioButton extends React.PureComponent<IProps> {
+    constructor(props: IProps) {
+        super(props);
+        this.handleChange = this.handleChange.bind(this);
```
petrjasek: fyi you don't have to bind if you call it via:
```ts
() => this.handleChannge(item.value)
```
only if you would pass it somewhere:
```ts
handleChange(event) {...}

<button onClick={this.handleChange}>
```

PR #407, opened 2019-12-10
feat(react): tab component

gagily:


app-typescript/components/TabList.tsx
```diff
@@ -0,0 +1,68 @@
...
+    theme?: 'light' | 'dark';
+    children: Array<any>;
+}
+
+interface IState {
+    index: number;
+}
+
+interface ITab {
+    text: string;
```
tomaskikutis: `label` would be a more specific name.

app-typescript/components/TabList.tsx
```diff
@@ -0,0 +1,68 @@
...
+class TabList extends React.PureComponent<ITabList, IState> {
+    constructor(props: ITabList) {
+        super(props);
+        this.state = {
+            index: 0,
+        };
+        this.handleChange = this.handleChange.bind(this);
+    }
+
+    handleChange(index: number) {
```
tomaskikutis: Can you make this `private` and add a `public` method `goTo(label: string)` which would enable a tab given a label. It would be called by using a ref.

app-typescript/components/TabList.tsx
```diff
@@ -32,14 +32,22 @@ class TabList extends React.PureComponent<ITabList, IState> {
...
     }
 
-    handleChange(index: number) {
+    private handleChange(index: number) {
         this.setState({
             index: index,
         });
     }
 
+    goTo(label: string) {
```
tomaskikutis: add a `public` keyword so it's clear for other people reading the component that it's meant to be exposed.

PR #409, opened 2019-12-11
feat(react): left navigation component

gagily:


tomaskikutis:
I would remove absolute positioning and separate the black toolbar from the menu component.

For the items list, I'd also use a structure where groups of items are passed as objects so it's easier to use as well as the compiler would be able to verify it better.

```
interface IMenuItem {
    id: string;
    label: string;
}

interface IMenuGroup {
    label: string;
    items: Array<IMenuItem>;
}

interface IMenu {
    groups: Array<IMenuGroup>;
    activeItemId: string;
    onSelect(id: string): void;
}

class Menu extends React.PureComponent<IMenu> {
...
}
```

tomaskikutis:
When I was asking to remove absolute positioning, I meant [this](https://user-images.githubusercontent.com/2076953/70895016-70996500-1fee-11ea-9d02-5c66cd0a6bc5.png).

Can you also extract that top menu bar as a separate component too?

Don't forget not to rebase the branch :wink:

pavlovicnemanja:
@tomaskikutis why not rebasing PR? In this case we are adding and fixing conflicts on currently opened PR.

tomaskikutis:
You can fix the conflicts as easily by merging(it's even easier since you only have one step). Rebasing the branch rewrites git history. When I review the same PR second or 3rd time, I only review the new commits. When you rewrite the history, there might as well be changes in the commits I have already reviewed, even if by mistake, and I have to review them again.

PR #417, opened 2019-12-23
feat(react): sliding toolbar, strech bar and sub nav components

gagily:


tomaskikutis:
I'm not [too impressed](https://user-images.githubusercontent.com/2076953/71371817-c18ef780-25b2-11ea-8a7c-4ef76ffa4546.png) to be honest :smile:

I would suggest better interfaces, but since there are no examples, I'm not sure what are the components supposed to do.

I don't see any exports, why? If you want to have some experimental components for designers to play around, it would be better if you created another branch and set them up there so you can experiment without having that code go through review and having to conform to production standards. When you are confident about those components, you could then refactor them, fix code quality add exports and submit a PR to the master branch.

PR #446, opened 2020-03-24
React dropdown

gagily:


tomaskikutis:
I suggest using an interface based on plain JS objects rather than on composition of React components because the former can be type-checked by a compiler which in my opinion also makes life easier for consumers.

I'd drop `side` and `align` from the interface and have auto-positioning implemented. I suggest replacing `navDropdown` with a method to pass any custom trigger element, not only the one specific to the navigation component.

Here's an interface I think would work better. I also added an example implementing [this](https://user-images.githubusercontent.com/2076953/81832268-98a9d780-953e-11ea-8222-756fb5a5819f.png) menu. I'm calling it `Menu` in my example, which isn't that important, except for being clear that this is not an input component, but something more like a [menu in the OS](https://user-images.githubusercontent.com/2076953/81833097-9431ee80-953f-11ea-96c8-2e7af5b43d5e.png).


```
interface IMenuItem {
    label: string;
    icon?: string;
    onSelect(): void;
}

interface IMenuGroup {
    groupLabel?: string;
    items: Array<IMenuItem | IMenuGroup>;
}

interface IMenu {
    label?: string;
    items: Array<IMenuItem | IMenuGroup>;
    header?: Array<IMenuItem | IMenuGroup>;
    footer?: Array<IMenuItem | IMenuGroup>;
    toggleElement: React.ComponentType<{toggle: () => void;}>;
}

export class Menu extends React.Component<IMenu> {/* implementation */}


//

function example() {
    return (
        <Menu
            header={[
                {
                    groupLabel: 'Create new item',
                    items: [
                        {label: 'editor 3 full', icon: 'plus', onSelect: () => 1},
                    ],
                }
            ]}
            items={[
                {
                    groupLabel: 'Recent templates',
                    items: [
                        {label: 'template 1', icon: 'plus', onSelect: () => 1},
                        {label: 'template 2', icon: 'plus', onSelect: () => 1},
                        {label: 'template 3', icon: 'plus', onSelect: () => 1},
                        {label: 'template 4', icon: 'plus', onSelect: () => 1},
                        {label: 'template 5', icon: 'plus', onSelect: () => 1},
                        {label: 'More templates...', onSelect: () => 1},
                    ],
                }
            ]}
            footer={[
                {
                    items: [
                        {label: 'Create Package', icon: 'package', onSelect: () => 1},
                        {label: 'Upload Media', icon: 'upload', onSelect: () => 1},
                    ],
                }
            ]}
            toggleElement={({toggle}) => <button className="icon-plus-big" onClick={() => toggle()}>+</button>}
        />
    );
}
```

What do you think @petrjasek @pablopunk ?

petrjasek:
yep I would prefer it to be data driven and keep those extra components internal. not sure if that toggle element should be part of it, maybe just a reference for position. and we need at least in sd to append dropdowns to body element like modals, so it gets displayed over areas with `overflow: hidden`.

tomaskikutis:
> not sure if that toggle element should be part of it

Some menus are triggered by an icon button, others by a select-like button with a label - [image](https://user-images.githubusercontent.com/2076953/81843723-4bcdfd00-954e-11ea-8849-0bf687cc3927.png). There might be some other types too. How would you do this without having a parameter for an element? Have multiple components for each type like MenuIcon and MenuLabel?

> need at least in sd to append dropdowns to body element like modals, so it gets displayed over areas with overflow: hidden.

I agree, but this should be an implementation detail, no need for the interface to know about it. All menus can be attached to the body by default. If we pass that `toggleElement`, in the implementation side it can be wrapped in a div, and then it can be positioned according to the ref of the wrapper.

petrjasek:
> > not sure if that toggle element should be part of it
> 
> Some menus are triggered by an icon button, others by a select-like button with a label - [image](https://user-images.githubusercontent.com/2076953/81843723-4bcdfd00-954e-11ea-8849-0bf687cc3927.png). There might be some other types too. How would you do this without having a parameter for an element? Have multiple components for each type like MenuIcon and MenuLabel?
> 

what about

```ts
<Dropdown ...>
    <i className="some btn"></i>
</Dropdown>
```

> > need at least in sd to append dropdowns to body element like modals, so it gets displayed over areas with overflow: hidden.
> 
> I agree, but this should be an implementation detail, no need for the interface to know about it. All menus can be attached to the body by default. If we pass that `toggleElement`, in the implementation side it can be wrapped in a div, and then it can be positioned according to the ref of the wrapper.

right, I'm just saying that because I don't see that detail to be implemented in this PR

tomaskikutis:
```
<Dropdown ...>
    <i className="some btn"></i>
</Dropdown>
```

This would work too, except it's a little bit less flexible, because you would have a button element on the UI framework side which you wouldn't be able to control if you needed to add a classname or wrap that button in another element etc.

tomaskikutis:
@gagily here's the interface we agreed on with Petr. 

```
interface IMenuItem {
    label: string;
    icon?: string;
    onSelect(): void;
}
interface IMenuGroup {
    groupLabel?: string;
    items: Array<IMenuItem | IMenuGroup | 'divider'>;
}
interface IMenu {
    label?: string;
    items: Array<IMenuItem | IMenuGroup | 'divider'>;
    header?: Array<IMenuItem | IMenuGroup | 'divider'>;
    footer?: Array<IMenuItem | IMenuGroup | 'divider'>;
}
export class Menu extends React.Component<IMenu> {/* implementation */}
```

What has changed from the initial one is the `divider` was added and also `toggleElement` was removed from the interface. The toggle element would be passed as children as Petr suggested in the comment above:

```
<Dropdown ...>
    <i className="some btn"></i>
</Dropdown>
```

tomaskikutis:
I've tried out the code and found some issues. Mainly with positioning. Some of the issues that are related to submenus aren't even worth fixing in my opinion, because when you implement submenus which attach to the body, you would throw away the current positioning code anyway.

[1 - styles](https://user-images.githubusercontent.com/2076953/82458078-8ab40380-9ab6-11ea-86aa-4749af80341d.png)
[2 - styles expanded](https://user-images.githubusercontent.com/2076953/82458080-8be53080-9ab6-11ea-95ea-055a584e0dfd.png)
[3 - no-body](https://user-images.githubusercontent.com/2076953/82458083-8c7dc700-9ab6-11ea-922c-6cc805169f21.png)
[4 - positioning top](https://user-images.githubusercontent.com/2076953/82458085-8c7dc700-9ab6-11ea-8d76-7979a2e9f261.png)
[6 - submenus](https://user-images.githubusercontent.com/2076953/82458087-8d165d80-9ab6-11ea-9cd6-50b34c6214e9.png)
[7 - submenus inside scrollable body](https://user-images.githubusercontent.com/2076953/82458088-8daef400-9ab6-11ea-9684-c32fc5f0e30e.png)
[8 - positioning bottom](https://user-images.githubusercontent.com/2076953/82458089-8e478a80-9ab6-11ea-9800-6d8e0fef999f.png)

app-typescript/components/Dropdown.tsx
```diff
@@ -0,0 +1,270 @@
...
+    const ref = React.useRef(null);
+    let inDebounce = 0;
+
+    React.useLayoutEffect(() => {
+        let element = document.getElementsByClassName('dropdown')[0];
+        let parentElement = getScrollParent(element);
+        function applyScroll() {
+            return change ? debounce(50) : null;
+        }
+        parentElement.parentNode.addEventListener("scroll", applyScroll());
```
tomaskikutis, line 53: what is `applyScroll` for? Shouldn't scrolling work natively?

app-typescript/components/Dropdown.tsx
```diff
@@ -0,0 +1,270 @@
...
+            clearTimeout(inDebounce);
+        };
+    }, [open]);
+
+    React.useEffect(() => {
+        setChange(true);
+    }, [open]);
+
+    // open close
+    function isOpen() {
```
tomaskikutis, line 68: A better name would help. Maybe `toggleDisplay` or the same `setOpen` you are already using. You could change the existing one to `setOpenState` for example to avoid collisions. is + adjective naming is mostly used to read variables and return a boolean.

app-typescript/components/Dropdown.tsx
```diff
@@ -0,0 +1,270 @@
...
+
+    const classes = classNames('dropdown', {
+        ['open']: open,
+        ['dropdown--dropup']: heightSet(height),
+        ['dropdown--align-right']: elementAlign() === 'right',
+    });
+
+    return (
+        <div className={classes} >
+            {typeof children === 'object' ?
```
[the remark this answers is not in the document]
gagily: <Dropdown ...>
    <i className="some btn"></i>
</Dropdown> 
In this case it is true, and in this case
<Dropdown ...>
    Button name
</Dropdown> 
it's not. In case you want default style of button you just write 'string' as a child of Dropdown component and if you want a styled button for example like navigation button then as a child of component send that navigation button as the element.

app-typescript/components/Dropdown.tsx
gagily: applyScroll() function is for calculating the position of an element on scroll only when the dropdown is open. I'll rename the function for better understanding, now it's confusing. :)

tomaskikutis: I wasn't referring to naming this time. What are you debouncing and why do you need an event listener on the `scroll` event?

pavlovicnemanja:
hey @tomaskikutis can we merge this?

tomaskikutis:
You are the boss now Nemaja :wink: The interface is fine, so check if it's working and if you are happy with the code and go ahead with the merging.

PR #450, opened 2020-03-30
Restore avatar styles used by planning

tomaskikutis:
SDBELGA-292

### DIFF
--- a/app/styles/components/_list-item.scss
+++ b/app/styles/components/_list-item.scss
@@ -404,6 +404,47 @@ $sd-ListItem-column-border: scale-color($grayNeutral, $alpha: -82%);
     }
 }
 
+
+.avatar {
+    display: inline-block;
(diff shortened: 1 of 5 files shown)


app/styles/components/_list-item.scss
```diff
@@ -404,6 +404,47 @@ $sd-ListItem-column-border: scale-color($grayNeutral, $alpha: -82%);
     }
 }
 
+
+.avatar {
```
pavlovicnemanja: You don't need to change this to `.sd-avatar`?

tomaskikutis: Nope. `sd-avatar` classnames are only for a react component. I'm keeping this for compatibility where react component isn't used.

pavlovicnemanja: 👍

PR #531, opened 2020-12-23
Add dropdown from prime react

tomaskikutis:
SDESK-5692

Also updated primereact version to 5.0.2(new API was added that I needed for the dropdown). I have done some fixes for calendar styles that seems to have changed.

@pavlovicnemanja can you check if your donut chart component works the same?


tomaskikutis:
@pavlovicnemanja seems like netlify thing is broken. Could you take a look?

pavlovicnemanja (approved):

PR #556, opened 2021-05-27
simplify carousel api

petrjasek:


app-typescript/components/Carousel.tsx
```diff
@@ -1,21 +1,28 @@
...
     autoplayInterval?: number;
     numVisible?: number;
     numScroll?: number;
     responsiveOptions?: Array<IPropsResponsive>;
-    headerTemplate?: JSX.Element;
-    footerTemplate?: JSX.Element;
-    itemTemplate(item: any): JSX.Element;
     onPageChange?(e: {page: number}): void;
+    title?: string;
+    imageCount?: number;
```
tomaskikutis, line 23: Why not use `images.length`?

I'd be good to reorder the props, so the mandatory ones are at the top.

tomaskikutis: I see. Can you add comments to the interface so it's clear how to use it? I assume if that prop is not passed, it would use `images.length`.

PR #558, opened 2021-06-10
Menu component

tomaskikutis:
SDCP-545


examples/pages/react/Menu.tsx
```diff
@@ -0,0 +1,121 @@
...
+            <section className='docs-page__container'>
+
+                <h2 className='docs-page__h2'>Menu</h2>
+
+                <p className='docs-page__paragraph'></p>
+                <Markup.ReactMarkup>
+                    <Markup.ReactMarkupPreview>
+                        <div className='docs-page__content-row docs-page__content-row--no-margin'>
+                            <div className='form__row'>
+                                <input type="text" />
```
pavlovicnemanja, line 73: I think this is not needed.

examples/pages/react/Menu.tsx
```diff
@@ -0,0 +1,121 @@
...
+        label: 'Spike item',
+        icon: 'icon-trash',
+        onClick: () => console.log('action: Spike item'),
+    },
+];
+
+export class MenuDocs extends React.Component {
+    render() {
+        return (
+            <section className='docs-page__container'>
```
pavlovicnemanja, line 64: Can you add minimal example code at top like on other pages?

image

Also include small description and props list?

package.json
```diff
@@ -82,12 +82,12 @@
   "dependencies": {
     "@material-ui/lab": "^4.0.0-alpha.56",
     "@popperjs/core": "^2.4.0",
+    "@superdesk/primereact": "^5.0.2-3",
```
pavlovicnemanja, line 85: Is it safe to go to version 6x?

examples/pages/react/Menu.tsx
tomaskikutis: I'll remove it. I initially had it in order to test the focus easily.

examples/pages/react/Menu.tsx
tomaskikutis: I have added example code in the markup section.

Regarding props, there is an interface defined that specifies all possible configurations. Why rewrite the same info in the docs? Whoever is going to use a component can jump to the interface in a single click. I'm happy to add comments to interface properties if you find some unclear.

go-to-interface

package.json
tomaskikutis: I don't think so. It's a major release so we would need to re-test all components that we use.

examples/pages/react/Menu.tsx
pavlovicnemanja: I agree, but we also use the props list so developers can easily see if component covers everything they need. All major frameworks use this approach so we want to have it also. All over components have it.

tomaskikutis: Doesn't jumping to the interface allow to easily see the same info? :smile:

Regarding other frameworks, we can borrow good practices that are useful for us, but we don't need to copy everything. For example, a lot of frameworks are written in js and don't have interfaces, so documenting props makes sense for them. We do have interfaces, so we aren't getting much value from describing the same info in a different way. It's more of a maintenance burden for us.

@petrjasek what do you think?

pavlovicnemanja: In that case you need to have imported that component to see if there is functionality you need. If you have everything listed, you can just go through components and see which one suits you the best. This is how we work, so please let's keep it consistant.

fritzSF: Come on @tomaskikutis, please add the damn list. I don't see what the big deal is, it won't take much of your time. It'll probably take less effort than going into a discussion about it. I get that you or other developers might not need it but this is not just for developers. Non-developers, I included, can use the components to create more complex layouts, playground pages etc. It's good to have an overview without having to dig into the code. Please, let's keep the practice we started.

pavlovicnemanja (approved):
LGTM! Just fix conflicts and merge

PR #559, opened 2021-06-10
Tooltip: appendToBody option

tomaszrondio:


fritzSF:
Needs a bit of polishing of the styles but it's good to go. I'll cover this part later.

fritzSF (approved):

PR #621, opened 2022-05-19
feat(TreeSelect): added props allowMultiple

dzonidoo:


tomaskikutis:
@dzonidoo I've done some testing, here are the issues I've found
* type definitions for `onChange` are incorrect and values can't be accessed without TypeScript throwing errors. See examples.
* Option template defined ambiguous types `ITreeNode<T> | T`. Is it `T` or `ITreeNode<T>`?
* `onChange` returns inconsistent object shape when component has initial value and another one is added: `[{"qcode":"editor","name":"Editor"},{"value":{"qcode":"writer","name":"Writer"}}]`
* `ITreeNode` should be exported from ui framework
* options that are already selected should not be selectable again

I've pushed 2 examples to your branch, make sure they compile and run. You can adjust the examples if needed, but don't disable type checking anywhere.

PR #647, opened 2022-06-28
bugfix - clicking on Menu inside actions also triggers click on Boxed ListItem

tomaskikutis:


app-typescript/components/Menu.tsx
```diff
@@ -60,6 +60,11 @@ interface IProps {
 
 const superdeskTopBarZIndex = 1030;
 
+interface ITieredMenuEvent {
+    item: any;
```
pavlovicnemanja: can we use some more specific type here?

pavlovicnemanja (approved):

PR #697, opened 2022-10-17
fix(TableList): fixed functionality for adding item.

dzonidoo:


app-typescript/components/Lists/TableList.tsx
```diff
@@ -208,7 +210,7 @@ export interface IPropsItem {
     end?: React.ReactNode;
     action?: React.ReactNode;
     addItem?: boolean;
-    itemsDropdown?: any;
+    itemsDropdown?(index?: any): any;
```
pavlovicnemanja: We should avoid using `any` here, in both parameter and return statement

pavlovicnemanja (approved):

PR #733, opened 2023-01-09
Pagination 3.0.0

thecalcc:


pavlovicnemanja (approved):

app-typescript/components/with-pagination.spec.tsx
```diff
@@ -0,0 +1,115 @@
...
+}
+
+// Simulate fetching delay
+const TIMEOUT = 1000;
+
+export class Paginated extends React.PureComponent {
+    getItems(): Promise<{items: Array<IPost>, itemCount: number}> {
+        return new Promise((resolve) => {
+            setTimeout(() => {
+                return resolve({items: range(1, 500).map((x) => ({title: `title ${x}`})), itemCount: 100});
```
tomaskikutis: shouldn't `itemCount` be 500 if you have 500 items?

app-typescript/components/with-pagination.spec.tsx
```diff
@@ -0,0 +1,115 @@
...
+            >
+                {
+                    (items) => <div>{JSON.stringify(items)}</div>
+                }
+            </WithPagination>
+        );
+    }
+}
+
+describe.only('with-pagination', () => {
```
tomaskikutis: `describe.only` is only to be used in development

app-typescript/components/WithPagination.tsx
```diff
@@ -0,0 +1,227 @@
+import * as React from 'react';
+import {Icon} from '../components/Icon';
+
+interface IProps<T> {
+    pageSize?: number;
```
tomaskikutis: it's best to put optional properties at the bottom

app-typescript/components/WithPagination.tsx
```diff
@@ -0,0 +1,227 @@
...
+            this.abortController.abort();
+        }
+
+        this.inProgress = true;
+        this.props.getItems(page, this.abortController.signal).then((res) => {
+            this.inProgress = false;
+            this.setState({items: res.items, currentPage: page}, () => {
+                const scrollableEl = getScrollParent(this.ref);
+                const diff = scrollableEl != null && this.ref?.scrollHeight != null
+                    ? scrollableEl.offsetHeight - this.ref?.scrollHeight
```
tomaskikutis: > scrollableEl.offsetHeight - this.ref?.scrollHeight

This might evaluate to `200 - undefined` - and we should make sure it doesn't happen. Actually, you do make sure, by only executing it if `this.ref?.scrollHeight != null`. But then you shouldn't be using optional chaining.

app-typescript/components/WithPagination.tsx
```diff
@@ -0,0 +1,227 @@
...
+
+    render() {
+        if (this.state.items == null) {
+            return null;
+        }
+
+        const pageElements = getPagination(this.state.currentPage, this.pageCount).map((el, i) => {
+            if (el === 'dots') {
+                return (
+                    <span data-test-id="span" className='sd-pagination__item sd-pagination__item--more'>...</span>
```
tomaskikutis: For data-test-id you should use names that make it easy to guess which UI element you're referring to.

examples/pages/react/WithPaginationDocs.tsx
```diff
@@ -0,0 +1,57 @@
...
+                    </Markup.ReactMarkupPreview>
+                    <Markup.ReactMarkupCode>{`
+                        <WithPagination
+                            getItems={(pageNo) => getItems(pageNo)}
+                        >
+                            {
+                                (items) =>
+                                    <div
+                                        style={{
+                                            display: 'flex',
```
tomaskikutis: In docs you'd want to avoid clutter and only show essential usage info. Styles here are that clutter that doesn't add anything to usage information.

app-typescript/components/with-pagination.spec.tsx
```diff
@@ -0,0 +1,115 @@
...
+    it('returns button1 when rendered', (done) => {
```
tomaskikutis: what is button 1 and why should it be rendered? You should do a better job via naming to indicate this. For one, I suggest testing `getPagination` function separately. Then what remains to test in the UI component is to do a simple test for one scenerio to see whether pagination bar renders. Then test behavior - next,prev - specific page and scrolling.

app-typescript/components/WithPagination.tsx
thecalcc: `ref` property is defined as `HTMLDivElement | null` so I have to do optional chaining as I assign the ref to null in the constructor.

examples/pages/react/WithPaginationDocs.tsx
thecalcc: I thought this should be accurate to represent the actual code of how the given example is implemented. Also I don't think it's clutter as it shows the functionality that you can style the paginated output however you like it.

app-typescript/components/WithPagination.tsx
tomaskikutis: you have to do optional chaining on line 110, but not on line 111. Line 11 only executes if it's not `null`

examples/pages/react/WithPaginationDocs.tsx
tomaskikutis: Even if that `div` was without styles - it would still be clear that you can style it, the same way it's clear that you can add other attributes, event listeners etc. even if they're not in the example - if you show that it's possible to use your element you make it clear then that everything is possible that you can do with an element.

app-typescript/components/with-pagination.spec.tsx
```diff
@@ -34,35 +34,68 @@ export class Paginated extends React.PureComponent {
...
+
+    it('returns dots twice when on the middle page', () => {
+        assert.strictEqual(getPagination(5, 10).filter((x) => x === 'dots').length, 2);
+    });
+
+    it('returns page 1 and last page when on the middle page', () => {
+        const res = getPagination(5, 10);
+        const includesFirstAndLast = res[0] === 1 && res[res.length - 1] === 10;
+        assert.strictEqual(includesFirstAndLast, true);
+    });
```
tomaskikutis: can you check the entire output?
`expect getPagination(5, 10) to be [1, 2, 3, '...']`  (pseudo code here)

PR #740, opened 2023-01-31
feat(splitter): Added splitter component

pavlovicnemanja:


tomaskikutis:
Can you describe how it works? What are the features? A screenshot or two would help as well.

pavlovicnemanja:
> Can you describe how it works? What are the features? A screenshot or two would help as well.

I'm currently preparing demo page with real use on playgrounds... will be there soon

pavlovicnemanja:
Demo page can be seen here: https://deploy-preview-740--inspiring-pasteur-fc7efb.netlify.app/#/playgrounds/react/rundown-splitter

tomaskikutis (commented):
Which 3rd party libraries did you consider? And why did you decide to go with a custom implementation instead of wrapping a 3rd party library?

***

I'll not ask you to rewrite this PR, but in the future use class based components instead of hooks. Regardless which is better, our team is much more skilled at maintaining class based ones.

***

Can you update the interface to this one? It'd be cleaner and more explicit how it works.

```
interface IProps {
    unit: 'px' | 'percent';
    targetPane: 'primary' | 'secondary';
    minSize: number; // for targetPane
    initialSize: number; // for targetPane
    onSizeChange?(primary: number, secondary: number): void;
    children: [React.ReactNode, React.ReactNode];
}
```

* It makes it explicit whether percent or pixels are used for values
* throws compilation time error if children count is not 2, because only 2 is supported
* `horizontal` prop is removed since it wasn't used anywhere
* min/initial size can only operate on one pane, `targetPane` enforces that at compile time preventing values for both panes to be passed at once. We also need to be able to specify initial secondary size which isn't possible using the current interface.
* `resetOnDoubleClick` prop I think we should drop and always apply the behavior. I see no reason why would we ever set it to false since usage is optional from end-user point of view.
* I've replaced `onSplitChanged` and `onMeasuredSizesChanged` with a single prop - `onSizeChange`. Let me know if you think I'm missing something. I couldn't understand why would we need both for the same thing.

pavlovicnemanja:
> Which 3rd party libraries did you consider? And why did you decide to go with a custom implementation instead of wrapping a 3rd party library?

It is created based on this component: https://primereact.org/splitter/ .
I think there are stuff we don't need, also applying styling to some 3rd party component will be much harder then creating our own with functionalities we need.

> I'll not ask you to rewrite this PR, but in the future use class based components instead of hooks. Regardless which is better, our team is much more skilled at maintaining class based ones.

Hooks are new things in React, and I think we should start using them because they have lot of benefits. I think we should go with time :)

> Can you update the interface to this one? It'd be cleaner and more explicit how it works.

Yep. I'm on it :)

pavlovicnemanja:
```
interface IProps {
    unit: 'px' | 'percent'; // ADDED
    targetPane: 'primary' | 'secondary'; // Target pane always needs to be primary, because of CSS grid which is used for calculation
    minSize: number; // minSize needs to be defined for both primary and secondary pane
    initialSize: number; // initialSize can be set only for primary target, again because of CSS grid used for handling panes
    onSizeChange?(primary: number, secondary: number): void;
    children: [React.ReactNode, React.ReactNode];
}
```

tomaskikutis (commented):
> Target pane always needs to be primary, because of CSS grid which is used for calculation

We need to adjust the implementation then. It's very limiting to only have one option. Also that's exactly the case in rundowns. The size of the right pane will be specified and the size of the left one will be what remains.

> minSize needs to be defined for both primary and secondary pane

Makes sense, let's have it for both.

> unit: 'px' | 'percent'; // ADDED

`onSizeChange` doesn't respect the unit

---

Here are a few bugs I found:

1. with  `unit="px" initialPrimarySize={100}`: when I click and hold the separator, before dragging starts, panes change size [[video](https://user-images.githubusercontent.com/2076953/230104412-28148813-f247-4e31-906b-d39813134aa8.mp4)]. It shouldn't happen.
2. with `unit="px" initialPrimarySize={100}`: dragging is really slow [[video](https://user-images.githubusercontent.com/2076953/230104412-28148813-f247-4e31-906b-d39813134aa8.mp4)]
3. with `unit="px" minPrimarySize={100} initialPrimarySize={100}` dragging doesn't work at all, pane sizes are fixed

tomaskikutis:
Closing in favor of https://github.com/superdesk/superdesk-ui-framework/commit/05c786302eb36055995b62c7cdc83df35acdf9da

I've wrapped "react-resizable-panels" library there. It doesn't support percent values out of the box, but we can always add an intermediatery component that'd be doing the conversion if we need to. Percent should be fine for now. Besides this limitation I didn't find any issues with how component works.

PR #744, opened 2023-03-08
Fix issue with selecting text

dzonidoo:
SDESK-6836

fix(TableList&ContentList): fixed issue with selecting text on rundown list and improve onCLick and onDoubleCLick function.

PR #745, opened 2023-03-09
Fixed title text selection on rundown list item 

dzonidoo:
SDESK-6836


tomaskikutis:
Why does this PR contain code from https://github.com/superdesk/superdesk-ui-framework/pull/744 ? Shouldn't it be one line change to fix the behavior with padding?

PR #747, opened 2023-03-21
Panel pinned

dzonidoo:
SDESK-6861

### DIFF
--- a/app-typescript/components/Layouts/Panel.tsx
+++ b/app-typescript/components/Layouts/Panel.tsx
@@ -1,8 +1,8 @@
 import * as React from 'react';
-import { Icon } from '../Icon';
 import { IconButton } from '../IconButton';
 import { Spinner, LoadingOverlay } from '../Spinner';
 import classNames from 'classnames';
+import {ButtonGroup} from '../ButtonGroup';
 
 // ============= Panel ============ //
 
@@ -47,11 +47,16 @@ interface IPropsPanelHeader {
     theme?: 'light-ui' | 'dark-ui'; // defaults to 'light
     className?: string;
     onClose?(): void;
+    onPin?(): void;
+    pinned?: boolean;
 }
 
-class PanelHeader extends React.PureComponent<IPropsPanelHeader> {
+class PanelHeader extends React.PureComponent<IPropsPanelHeader, {rotate: boolean}> {
     constructor(props: IPropsPanelHeader) {
         super(props);
+        this.state = {
+            rotate: this.props.pinned ? this.props.pinned : false,
+        };
     }
 
     render() {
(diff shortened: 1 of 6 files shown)


app-typescript/components/Layouts/Panel.tsx
```diff
@@ -47,11 +47,16 @@ interface IPropsPanelHeader {
     theme?: 'light-ui' | 'dark-ui'; // defaults to 'light
     className?: string;
     onClose?(): void;
+    onPin?(): void;
+    pinned?: boolean;
 }
 
-class PanelHeader extends React.PureComponent<IPropsPanelHeader> {
+class PanelHeader extends React.PureComponent<IPropsPanelHeader, {rotate: boolean}> {
```
thecalcc: Make an `IState` interface instead of inlining the rotate prop so we have consistency.

app-typescript/components/Layouts/Panel.tsx
```diff
@@ -61,22 +66,50 @@ class PanelHeader extends React.PureComponent<IPropsPanelHeader> {
...
+                            onClick={() => {
+                                if (this.props.onPin) {
+                                    this.props.onPin();
+                                    this.setState({
+                                        rotate: !this.state.rotate,
+                                    });
+                                }
+                            }}
+                            rotate={this.state.rotate ? '90' : undefined}
+                        />
```
thecalcc: On line 88 check if the onPin prop is not null like this `this.props.onPin != null`. 
On line 93 you don't need to check again if the onPin function is not null since you do that on line 88.
On line 100 check if rotate is not null like this `(this.state.rotate?.length ?? 0) > 0` 
Add brackets around the <IconButton /> component.
You can use `&&` instead of ternary.

app-typescript/components/Layouts/Panel.tsx
```diff
@@ -61,22 +66,50 @@ class PanelHeader extends React.PureComponent<IPropsPanelHeader> {
...
+                        : null
+                    }
+                    {this.props.onClose
+                        ? <IconButton
+                            icon='close-small'
+                            ariaValue='Close'
+                            onClick={this.props.onClose}
+                        />
+                        : null
```
thecalcc: Line 104 - check if the onClose function is not null explicitly. 
Add brackets around the `<IconButton />`
You can use `&&` here since the second option is `null`.

app-typescript/components/Layouts/Panel.tsx
```diff
@@ -95,15 +128,17 @@ class PanelContent extends React.PureComponent<IPropsPanelContent> {
...
+                    : null
                 }
-                {this.props.empty ?
-                    <LoadingOverlay>
+                {this.props.empty
+                    ? <LoadingOverlay>
                         {this.props.emptyTemplate}
-                    </LoadingOverlay> : null
+                    </LoadingOverlay>
+                    : null
```
thecalcc: No need for ternary, instead use `&&`. Put brackets around `LoadingOverlay`

tomaskikutis:
Closing this one. Another PR should be submitted that allows JSX inside panel header

PR #754, opened 2023-04-13
Avatar updates

fritzSF:


app-typescript/components/Avatar.tsx
```diff
@@ -12,7 +13,15 @@ interface IImageAvatar extends IBase {
 interface ITextAvatar extends IBase {
     text: string; // limited to 3 characters
 }
-
+interface INumberAvatar extends IBase {
+    number: string;
+}
+interface IIconAvatar extends IBase {
+    status?: 'default' | 'in-progress' | 'done';  // defaults to default
+    type?: 'text' | 'photo' | 'video' | 'audio' | 'file' | 'post' | 'post' | 'list-alt' | 'graphic';
```
MarkLark86, line 21: What is the benefit here of restricting what icons can be used with an Avatar?
I know we don't have other use cases right now, but would it be beneficial to allow any icon here (and remove the `canceled` prop, and the Planning module will choose which icon based on desired coverage assignment state)?

app-typescript/components/Avatar.tsx
```diff
@@ -12,7 +13,15 @@ interface IImageAvatar extends IBase {
 interface ITextAvatar extends IBase {
     text: string; // limited to 3 characters
 }
-
+interface INumberAvatar extends IBase {
+    number: string;
+}
+interface IIconAvatar extends IBase {
+    status?: 'default' | 'in-progress' | 'done';  // defaults to default
```
MarkLark86, line 20: Same with `status`, can be simply use the same `type` from the `Icon` component (and let the Planning module choose which colour to apply?)

tomaskikutis:
It should have been marked as WIP @MarkLark86. Fritz made the UI and I'll later review and fix the structure/interfaces.

tomaskikutis:
continued work in https://github.com/superdesk/superdesk-ui-framework/pull/758 I didn't push changes to this PR in order not to mess up Fritz's `develop` branch.

PR #757, opened 2023-04-25
Panel pinned

dzonidoo:


tomaskikutis (commented):
separate component for rotation
update example page

tomaskikutis (approved):

PR #758, opened 2023-04-28
Avatar improvements

tomaskikutis:
* Added standalone `Avatar` component. Keeping `AvatarWrapper` and children for backwards compatibility - will replace with standalone avatar in the future. Docs replaced to only refer to standalone component.
* Added avatar group
* Improved `Avatar` component so any icon/color combination can be used

based on https://github.com/superdesk/superdesk-ui-framework/pull/754

image

image


tomaskikutis:
@dzonidoo post comments here if something doesn't work when using this for updating planning UI

PR #765, opened 2023-05-10
Avatar group popover

dzonidoo:
SDESK-6925


tomaskikutis:
~~move `PlusButtonWrapper` to top level code (out of existing component)~~ improve interfave - drop children, inline props

tomaskikutis (approved):

PR #767, opened 2023-05-16
Introduced initially selected tab index

thecalcc:


thecalcc:
@dzonidoo there's one actual change - adding the initiallySelectedIndex prop and setting it in the `index` state initially. The rest of the changes are just formatting.

dzonidoo (approved):

PR #774, opened 2023-05-30
Improve time input

thecalcc:


app/styles/components/_theme-selector.scss
```diff
@@ -145,6 +145,10 @@
 }
 
 [data-theme="dark-ui"] {
+    input[type="time"]::-webkit-calendar-picker-indicator {
```
tomaskikutis, line 148: webkit is chromium, what about firefox? does it work there outside the box?

app/styles/form-elements/_inputs.scss
```diff
@@ -654,6 +654,7 @@
         grid-column: 2/4;
     }
     .sd-input__input-container {
+        min-width: 110px;
```
tomaskikutis: is this for all inputs? Shouldn't it only be for time picker?

app/styles/components/_theme-selector.scss
thecalcc: Firefox's native time input doesn't contain an icon like Chrome does so we don't need to think about that.

app/styles/form-elements/_inputs.scss
thecalcc: Yes, changed it so that we apply it to this class only if it contains a time input

tomaskikutis (approved):

PR #777, opened 2023-06-06
Fix issues after TreeSelect is unmounted

thecalcc:


dzonidoo (approved):

app-typescript/components/TreeSelect.tsx
```diff
@@ -129,38 +131,47 @@ export class TreeSelect<T> extends React.Component<IProps<T>, IState<T>> {
...
 
-    componentDidMount = () => {
-        this.recursion(this.state.options);
-        document.addEventListener("mousedown", (event) => {
-            if (
-                (this.dropdownRef.current?.contains(event.target as HTMLElement) !== true)
-                && (this.openDropdownRef.current?.contains(event.target as HTMLElement) !== true)
-            ) {
-                this.setState({openDropdown: false});
+    onMouseDown = (event: MouseEvent) => {
```
tomaskikutis, line 134: FYI this is not a class method, but a [public class field](https://github.com/tc39/proposal-class-fields) and doesn't need binding. It also works a bit differently than methods in more complex case like inheritance. It also has a  [negative performance impact](https://github.com/tc39/proposal-class-fields/issues/270) that might be significant or not depending on usage. Because of this and the fact that didn't fully investigate the feature I prefer using class methods. It might be worth looking into it in the future since it allows for shorter code in some cases.

PR #781, opened 2023-06-22
fixed button wrap issue

dzonidoo:
SDESK-6938

### DIFF
--- a/app/styles/form-elements/_checkbox.scss
+++ b/app/styles/form-elements/_checkbox.scss
@@ -416,6 +416,9 @@ $checkButtonBorderRadius: $border-radius__base--small;
     &.sd-check-button__group--compact {
         gap: $sd-base-increment * 0.5;
     }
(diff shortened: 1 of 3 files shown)


app/styles/form-elements/_checkbox.scss
```diff
@@ -416,6 +416,9 @@ $checkButtonBorderRadius: $border-radius__base--small;
     &.sd-check-button__group--compact {
         gap: $sd-base-increment * 0.5;
     }
+    &.sd-check-button__group--no-wrap {
```
tomaskikutis, line 419: does that classname that you're styling exist on some HTML element? I couldn't find it

dzonidoo: Yep but on client-core side

tomaskikutis (approved):
This is fine since that client-core markup is based on angular, but for react stuff, we shouldn't be using CSS directly from ui-framework - only components

PR #793, opened 2023-07-19
improve input components with error state

dzonidoo:
SDESK-7008

### DIFF
--- a/app-typescript/components/Input.tsx
+++ b/app-typescript/components/Input.tsx
@@ -60,8 +60,8 @@ export class Input extends React.Component<IProps> {
                 required={this.props.required}
                 disabled={this.props.disabled}
                 value={this.props.value}
-                invalid={this.props.error != null ? true : false}
                 error={this.props.error}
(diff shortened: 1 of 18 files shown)


tomaskikutis:
`const f = p != null` already produces a boolean result, there's no purpose in doing `p != null ? true : false`

app-typescript/components/Input.tsx
```diff
@@ -60,8 +60,8 @@ export class Input extends React.Component<IProps> {
                 required={this.props.required}
                 disabled={this.props.disabled}
                 value={this.props.value}
-                invalid={this.props.error != null ? true : false}
                 error={this.props.error}
+                invalid={this.props.error != null ? true : false}
```
tomaskikutis: maybe revert this if you're not actually changing anything?

dzonidoo: I just wanted to be neat 😄

tomaskikutis (approved):

PR #800, opened 2023-08-09
avatar changes required to replace planning avatars

tomaskikutis:
SDBELGA-739

### DIFF
--- a/app-typescript/components/avatar/avatar-image.tsx
+++ b/app-typescript/components/avatar/avatar-image.tsx
@@ -3,13 +3,17 @@ import {IPropsBase} from './interfaces';
 
 interface IPropsImageAvatar extends IPropsBase {
     imageUrl?: string | null; // defaults to a placeholder image
+    onClick?(): void;
 }
 
 export class AvatarContentImage extends React.PureComponent<IPropsImageAvatar> {
     render() {
-        if (this.props.imageUrl == null) {
+        const {imageUrl, tooltipText, onClick} = this.props;
+        const role: string | undefined = onClick == null ? undefined : 'button';
+
+        if (imageUrl == null) {
             return (
-                <span className="sd-avatar-content sd-avatar-content--dummy-img" title={this.props.tooltipText}>
+                <span className="sd-avatar-content sd-avatar-content--dummy-img" title={tooltipText} role={role}>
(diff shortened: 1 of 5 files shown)


app-typescript/components/avatar/avatar-placeholder.tsx
```diff
@@ -20,15 +21,35 @@ export interface IPropsAvatarPlaceholder {
...
+                    if (kind === 'plus-button') {
+                        return (
+                            <AvatarContentAdd
+                                tooltipText={tooltip ?? undefined}
+                                onClick={onClick}
+                            />
+                        );
+                    } else if (kind === 'user-icon') {
+                        return (
+                            <span onClick={onClick}>
```
MarkLark86: Interactive/clickable elements should probably be a `button` instead of a `span` (@fritzSF your thoughts?).

As we're adding click support to an `AvatarContentImage`, it would make sense to add this support into that component directly rather than for this single use case (just like we're already doing with the `AvatarContentAdd` component).

tomaskikutis: I've tried a button, but it was breaking styles, so I've used `role="button"` instead.

app-typescript/components/avatar/avatar-image.tsx
```diff
@@ -3,22 +3,26 @@ import {IPropsBase} from './interfaces';
...
 export class AvatarContentImage extends React.PureComponent<IPropsImageAvatar> {
     render() {
-        if (this.props.imageUrl == null) {
+        const {imageUrl, tooltipText, onClick} = this.props;
+        const role: string | undefined = onClick == null ? undefined : 'button';
+
+        if (imageUrl == null) {
             return (
-                <span className="sd-avatar-content sd-avatar-content--dummy-img" title={this.props.tooltipText}>
+                <span className="sd-avatar-content sd-avatar-content--dummy-img" title={tooltipText} role={role}>
```
MarkLark86, line 16: @tomaskikutis Where is the `onClick` property assigned?

tomaskikutis: good catch, somehow missed that. Fixing in https://github.com/superdesk/superdesk-ui-framework/pull/801

PR #804, opened 2023-08-21
Create TreeMenu component

dzonidoo:
SDESK-7040

### DIFF
--- a/app-typescript/components/TreeMenu.tsx
+++ b/app-typescript/components/TreeMenu.tsx
@@ -1,0 +1,448 @@
+import * as React from "react";
+import { Icon } from "./Icon";
+import { createPopper, Instance } from '@popperjs/core';
+import {getPrefixedItemId, TreeSelectItem} from './TreeSelect/TreeSelectItem';
+import {keyboardNavigation} from './TreeSelect/KeyboardNavigation';
+import {createPortal} from 'react-dom';
+
+interface IState<T> {
+    options: Array<ITreeMenuNode<T>>;
+    openDropdown: boolean;
+    activeTree: Array<Array<ITreeMenuNode<T>>>;
+    buttonTree: Array<ITreeMenuNode<T>>;
+    buttonValue: ITreeMenuNode<T> | null;
+    filterArr: Array<ITreeMenuNode<T>>;
+    searchFieldValue: string;
+    firstBranchOptions: Array<ITreeMenuNode<T>>;
+
+    // array of classNames; used to focus an item after returning for another level
+    buttonTarget: Array<string>;
+}
+
+interface IProps<T> {
+    width?: string;
+    zIndex?: number;
+    searchPlaceholder?: string;
+    singleLevelSearch?: boolean;
+    getOptions?(): Array<ITreeMenuNode<T>>;
+    getLabel(item: T): string;
+    getId(item: T): string;
+    getBackgroundColor?(item: T): string;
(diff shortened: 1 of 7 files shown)


app-typescript/components/TreeSelect/TreeSelectItem.tsx
```diff
@@ -25,7 +26,10 @@ export class TreeSelectItem<T> extends React.Component<IProps<T>> {
     render() {
         return (
             <li
-                className='suggestion-item suggestion-item--multi-select'
+                className={
+                    'suggestion-item suggestion-item--multi-select'
+                    + (this.props.labelWrap ? ' suggestion-item--wrap' : '')
```
tomaskikutis: how about making it wrap unconditionally?

app-typescript/components/TreeSelect/TreeSelectItem.tsx
```diff
@@ -55,7 +59,9 @@ export class TreeSelectItem<T> extends React.Component<IProps<T>> {
                     <span
                         className={
                             'suggestion-item--bgcolor'
-                            + (this.props.selectedItem ? ' suggestion-item--disabled' : '')}
+                            + (this.props.selectedItem ? ' suggestion-item--disabled' : '')
+                            + (this.props.labelWrap ? ' suggestion-item--bgcolor--wrap' : '')
```
tomaskikutis: is `bg-color` related to wrapping? how?

app-typescript/components/TreeMenu.tsx
```diff
@@ -0,0 +1,443 @@
...
+
+function nodeHasChildren<T>(item: IParent<T> | IChildren<T>): item is IParent<T> {
+    return item['children'] != null;
+}
+
+function nodeCanBeSelected<T>(item: IParent<T> | IChildren<T>): item is IChildren<T> {
+    return item['onSelect'] != null;
+}
+
+export class TreeMenu<T> extends React.Component<IProps<T>, IState<T>> {
```
tomaskikutis, line 56: I see there's lot of code duplication between this and tree select. I expect we should now be able to reimplement tree select something like this:

```

export class TreeSelect extends React.Component {
    render() {
        return (
            <TreeMenu
                getOptions={() => options2}
                getId={(item) => item.name}
                getLabel={(item) => item.name}
                optionTemplate={(item) => {
                    return (
                        <div
                            style={{
                                display: 'flex',
                                alignItems: 'center',
                                gap: '4px'
                            }}
                        >
                            {item.icon}
                            {item.name}
                        </div>
                    )
                }}
            >
                {(toggle) => (
                    <div>
                        <div>label</div>

                        <div>
                            {
                                this.state.values.map((val, i) => (
                                    <span key={i}>{val}</span>
                                ))
                            }
                        </div>

                        <button onClick={toggle}>+</button>
                    </div>
                )}
            </TreeMenu>
        );
    }
}
```

app-typescript/components/TreeSelect/TreeSelectItem.tsx
dzonidoo: but why?

app-typescript/components/TreeMenu.tsx
dzonidoo: but what about async mode on treeSelect?
treeMenu not support it

tomaskikutis (approved):

PR #806, opened 2023-09-14
create component TimePickerV2

dzonidoo:


thecalcc:
TimePickerV2.tsx: 
- remove unused imports
- hours, minutes, seconds - put them inside the class as private class properties, use range from `lodash` instead of manually crafting them.
- update handleChange function so it's like this 
`this.setState({[state]: event.target.value} as any); setTimeout(() => {this.props.onChange(this.state.value);})`
- extract the code that repeats 

```
if (this.props.allowSeconds) {
          this.setState({
              value: `${this.state.hours}:${this.state.minutes}:${this.state.seconds}`,
          });
      } else {
          this.setState({
              value: `${this.state.hours}:${this.state.minutes}`,
          });
      }
```

- use the short version inside `stringifyArray` 
- inside `stateUpdate` try thinking of a solution to avoid using `let`
- format code, so you have `this.props.allowSeconds && (` on one line
- inside the render method where you map state varables don't use index for key, instead use the varables themselves as keys since they're unqiue
- use shorthand for option

thecalcc:
You don't actually need to save the stringified values in the state. You can create three constants in the render method and stringify the respective values from props. This way you also won't need to do `setTimeout` to update the state. You'll only have to pass the new value to onChange to update the value in the parent component. After which this will trigger a rerender of your component and update the values.

app-typescript/components/TimePickerV2.tsx
```diff
@@ -0,0 +1,166 @@
...
+        if (!this.props.disabledOptions[timeUnit]?.includes(Number(value)) && value != null) {
+            return value;
+        } else {
+            return stringifyArr[0];
+        }
+    }
+
+    componentDidMount(): void {
+        if (this.props.allowSeconds) {
+            this.props.onChange(`${this.updateValue('hours', this.hoursStringified)}:${this.updateValue('minutes', this.minutesStringified)}:${this.updateValue('seconds', this.secondsStringified)}`);
```
tomaskikutis: I noticed linter doesn't catch too long lines in some cases. I found it's because of a linter version. @dzonidoo can you install `tslint@5.11.0` on develop (and fix new lint issue)?

tomaskikutis: forgot to mention to do this on develop, not this PR

tomaskikutis:
@thecalcc 

* avoid synchronizing prop, move class props to render method variables
* use native padstart
* updateValue, handleChangeForTimeUnit - rewrite into one(if possible) smaller function

tomaskikutis:
one more thing @thecalcc , in current implementation it is possible to choose `24:01` which shouldn't be possible

tomaskikutis:
or `24:15` I should say, the problem is with `24` though

thecalcc:
- is12HourFormat, make it a class property instead of a function, compute and assign it in the constructor, no need to compute it every time you use it.
- use `const` instead of `let`
- `update12HourValue` - call this `updated12HourValue`
- don't cast with `Number`, use `parseInt(value, 10)`
-  extract `padStart((Number(splitValue[0]) - 12).toString(), 2, '0')` to a function so you can pass to it just `splitValue[0]) - 12`
- remove `{` `}` in select options properties

thecalcc (approved):

tomaskikutis:
* remove implicit assumption that AM is default
* get rid of timeFormat - use value to determine if it's am or pm
* extract getting final timeUnitValuesArray to a separate function
* remove nested ternary

tomaskikutis:
@dzonidoo follow up with @thecalcc and do the changes in my comment

thecalcc (approved):
Did you verify that we can switch AM/PM from the select and when you do that the hours don't change?

PR #809, opened 2023-10-24
add fullWidth prop on TreeSelect

dzonidoo:
SDESK-7072


app-typescript/components/TreeSelect/TreeSelect.tsx
```diff
@@ -36,6 +36,7 @@ interface IPropsBase<T> extends IInputWrapper {
     selectBranchWithChildren?: boolean;
     readOnly?: boolean;
     width?: string;
+    fullWidth?: boolean;
```
tomaskikutis: why would we add another prop when we can pass `width: '100%'`?

tomaskikutis (approved):
don't forget to increase version number

tomaskikutis:
nevermid, I see you already did that

PR #815, opened 2024-01-24
Improve accessibility on tree component

dzonidoo:
SDUF-33


app-typescript/components/TreeMenu.tsx
```diff
@@ -11,7 +11,7 @@ interface IState<T> {
     openDropdown: boolean;
     activeTree: Array<Array<ITreeMenuNode<T>>>;
     buttonTree: Array<ITreeMenuNode<T>>;
-    buttonValue: ITreeMenuNode<T> | null;
```
tomaskikutis: what's wrong with `null`?

app-typescript/components/TreeMenu.tsx
```diff
@@ -428,13 +427,16 @@ export class TreeMenu<T> extends React.Component<IProps<T>, IState<T>> {
...
                                                 option={option}
                                                 handleTree={this.handleTree}
-                                                onClick={onSelect(option)}
-                                                disabledItem={disabledItem(option)}
-                                                getBorderColor={this.props.getBorderColor}
-                                                getBackgroundColor={this.props.getBackgroundColor}
                                                 getId={this.props.getId}
                                                 optionTemplate={this.props.optionTemplate}
                                                 getLabel={this.props.getLabel}
+                                                disabledItem={disabledItem(option)}
```
tomaskikutis: is there a reason you changed the order of props?

examples/pages/react/TreeMenu.tsx
```diff
@@ -176,6 +176,7 @@ export class TreeMenuDocs extends React.Component<{}, {}> {
                     <Markup.ReactMarkupPreview>
                         <div className='docs-page__content-row docs-page__content-row--no-margin'>
                             <TreeMenu
+                                searchPlaceholder='Search...'
```
tomaskikutis: We can't use english strings that would be visible to users. That's because translations wouldn't work then. Instead take placeholder as a prop.

dzonidoo: but this is just example page

app-typescript/components/TreeMenu.tsx
dzonidoo: actually nope...

app-typescript/components/TreeMenu.tsx
dzonidoo: i will show you

examples/pages/react/TreeMenu.tsx
tomaskikutis: good point, I didn't see this

tomaskikutis (approved):

PR #827, opened 2024-03-05
Select component fixes, SDESK-7182 examples

fritzSF:


tomaskikutis:
Where will `DropDownButton` be used?

fritzSF:
> Where will `DropDownButton` be used?

It is intended to be used in combination with the _Menu_ and/or _TreeMenu_ components, to replace the Dropdown. Any changes or improvemsnts on it can be done later as we won't use it right away. The focus ATM is on the the other files.

tomaskikutis:
The difference between `TreeSelect` and `TreeMenu` is that `TreeMenu` is for executing actions and doesn't have a value while `TreeSelect` is for selecting a value. This dropdown button has a value, so it would be rather for `TreeSelect`. `TreeSelect` does accept templates, but it accepts only the inner part - without the caret or borders. Since you say we wouldn't be using it right away, could you then remove the `DropDownButton` related parts from this PR? You can save it on some other branch and later submit it as a separate PR where we would get it 100% working before merging.

fritzSF:
> The difference between `TreeSelect` and `TreeMenu` is that `TreeMenu` is for executing actions and doesn't have a value while `TreeSelect` is for selecting a value. This dropdown button has a value, so it would be rather for `TreeSelect`. `TreeSelect` does accept templates, but it accepts only the inner part - without the caret or borders. Since you say we wouldn't be using it right away, could you then remove the `DropDownButton` related parts from this PR? You can save it on some other branch and later submit it as a separate PR where we would get it 100% working before merging.

OK, then it can be used with the `Menu`.  What is the harm keeping it there for now? It's a small component

tomaskikutis:
It doesn't make sense to use it with `Menu` since `Menu` is for executing actions and doesn't have a value.

The harm is maintanence cost. If something doesn't compile for example or we're renaming CSS properties - we do it also for unused code - because we assume that if it was merged - it must be used. To go over usages would be even more work. Also the more code there is - the harder is to find what you need. It's especially annoying to find that something was added to the code that never worked.

fritzSF:
Okay, let's not discuss it here. The fact is that we need a component to replace the DropDown. The button part should contain all the options that the new DropDownButton component has (with more to be added later); the dropdown part should have the same functionality as the current 'Menu' component. What we will use in the end is not important.

I will remove the DropDown from the PR now, although I still think it's a waste of time.

fritzSF:
Done

tomaskikutis (approved):

PR #831, opened 2024-04-08
abstract label in authoring should have the same style as all other labels

dzonidoo:
SDESK-7233


tomaskikutis (approved):

tomaskikutis:
in general it would be a bad approach since ui-framework should not know about its usages and classnames outside of it, but we'll probably remove all authoring css from here after we drop angular based authoring so let's save time and keep it for now

PR #843, opened 2024-04-18
TreeSelect - add props openedDropdown

dzonidoo:


app-typescript/components/TreeSelect/TreeSelect.tsx
```diff
@@ -42,6 +42,8 @@ interface IPropsBase<T> extends IInputWrapper {
     singleLevelSearch?: boolean;
     placeholder?: string;
     searchPlaceholder?: string;
+    noResultsFoundMessage?: string;
+    openedDropdown?: boolean;
```
tomaskikutis, line 46: update naming to `dropdownInitiallyOpen` to clarify that the prop applies only to first render.

app-typescript/components/TreeSelect/TreeSelect.tsx
```diff
@@ -42,6 +42,8 @@ interface IPropsBase<T> extends IInputWrapper {
     singleLevelSearch?: boolean;
     placeholder?: string;
     searchPlaceholder?: string;
+    noResultsFoundMessage?: string;
```
tomaskikutis, line 45: make this one mandatory and do not provide a default value - it can't be translated.

dzonidoo: are u sure about mandatory?

tomaskikutis: good point, many usages would break, let's do it another time

dzonidoo:
duplicate off https://github.com/superdesk/superdesk-ui-framework/pull/850

PR #845, opened 2024-04-22
Tree menu focused element issue

dzonidoo:


thecalcc (approved):

tomaskikutis:
@thecalcc you should check that ticket number is added in the description. How should I find from this PR what was wrong and what's the idea with this change?

PR #860, opened 2024-06-06
Subnav css fix

dzonidoo:


tomaskikutis:
> Fix subnav for publisher

It shouldn't matter where the component is used and fixes should be generic. Is this specific in any way to publisher or you simply didn't pick a good name for the PR?

dzonidoo:
the name of pr is just inherited from the name of commit and I didnt notice, which is my mistake yep...

tomaskikutis (approved):

PR #861, opened 2024-06-07
The DatePicker in the Rundown section should stretch to the available width

dzonidoo:
SDESK-7301


tomaskikutis:
How does date picker know if it's being rendered in the `rundowns` section? :upside_down_face:

Do not use language referring to component usages outside of this project. That's not relevant. What is relevant is features you're adding to components and how it impacts all existing usages. Are you adding a behaviour that didn't exist? Or are you changing existing behavior? Will it change for all components or is it configurable?

PR #863, opened 2024-06-12
Create a new CollapseBox component

dzonidoo:
SDUF-110

### DIFF
--- a/app-typescript/components/ToggleBox/index.tsx
+++ b/app-typescript/components/ToggleBox/index.tsx
@@ -2,7 +2,7 @@ import * as React from 'react';
 import {SimpleToggleBox} from './SimpleToggleBox';
 import {CustomHeaderToggleBox} from './CustomHeaderToggleBox';
 
-export interface IPropsWithLabel {
+export interface IPropsSimple {
     variant: 'simple';
     title: string;
     badge?: JSX.Element;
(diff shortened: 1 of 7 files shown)


app-typescript/components/ToggleBox/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+    initiallyOpen?: boolean; // defaults to false
+    className?: string;
+    margin?: 'none' | 'small' | 'normal' | 'large';
+    onOpen?(): void;
+    onClose?(): void;
+}
+
+export interface IPropsWithoutLabel {
+    variant: 'custom-header';
+    // always visible
```
tomaskikutis: image

use line breaks to make it more clear which line a comment applies. In this case, since the comment is short it's even better to put it on the same line.

app-typescript/components/ToggleBox/index.tsx
```diff
@@ -0,0 +1,42 @@
+import * as React from 'react';
+import {SimpleToggleBox} from './SimpleToggleBox';
+import {CustomHeaderToggleBox} from './CustomHeaderToggleBox';
+
+export interface IPropsWithLabel {
```
tomaskikutis: interface naming should match the variant. If variant is `simple`, interface should be `IPropsSimple`

app-typescript/components/ToggleBox/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+    onOpen?(): void;
+    onClose?(): void;
+}
+
+export interface IPropsWithoutLabel {
+    variant: 'custom-header';
+    // always visible
+    header: React.ReactNode;
+    children: React.ReactNode;
+    toggleButton: string;
```
tomaskikutis: call this `toggleButtonLabel`. It's not the entire button that's passed as prop.

tomaskikutis (commented):
`npm start` doesn't work due to a bad export. IT should be updated to

image

tomaskikutis (approved):

PR #866, opened 2024-07-03
Add clear button datepicker

thecalcc:


thecalcc:
Just to confirm you're ok if we add the clear button to show by default?

thecalcc:
This is how it looks:

image

thecalcc:
This would actually revert the functionality of existing DatePickers which have no "Clear" functionality existing. We'd rather have to do a `notRequired` prop to preserve existing functionality.

Also think about cases like the Publishing panel for articles, where we have clear functionality separated, and it links a DatePicker field with a TimePicker field, to reset them both

app-typescript/components/DatePicker.tsx
```diff
@@ -39,7 +40,7 @@ interface IDatePicker extends IDatePickerBase {
     onChange(valueNext: Date | null): void;
     maxDate?: Date;
     minDate?: Date;
-    showButtonBar?: boolean;
+    required?: boolean;
```
tomaskikutis: `required` prop already exists from extended interfaces

tomaskikutis:
> Just to confirm you're ok if we add the clear button to show by default?

Sure if it won't break anything.

https://github.com/superdesk/superdesk-ui-framework/assets/2076953/099b00a6-f040-44c6-8efa-aae30afdfaeb

It looks like it already allows to clear values manually anyway so a clear button shouldn't break anything. Especially if you make sure to pass the same value(maybe it should be empty string instead of `null`) in `onChange` as being passed if you simply clear the field with `ctrl+a` and `backspace`.

fritzSF:
@thecalcc It would be better not to affect the current UX and have the clear only as an option. Also, please align the action to the right, this would be in line with our general standard (e.g. modals, dialogs etc.)

tomaskikutis:
> Also think about cases like the Publishing panel for articles, where we have clear functionality separated, and it links a DatePicker field with a TimePicker field, to reset them both

We could do `required={true}` there. But even if we don't it won't crash. It would only improve UI since that field can be also cleared independently manually anyway.

> We'd rather have to do a notRequired prop to preserve existing functionality.

We should avoid negatives in naming. It'd be better then to do `allowClearing` or go over usages and pass `required={true}` where needed. But I think we should be good without neither of those.

tomaskikutis:
> It would be better not to affect the current UX

Why not @fritzSF ? If we're affecting it in a good way? :) It's not a text input, so clearing by removing invidivual characters that we have by default now is not good UX to begin with

thecalcc:
> > Just to confirm you're ok if we add the clear button to show by default?
> 
> 
> 
> Sure if it won't break anything.
> 
> 
> 
> https://github.com/superdesk/superdesk-ui-framework/assets/2076953/099b00a6-f040-44c6-8efa-aae30afdfaeb
> 
> 
> 
> It looks like it already allows to clear values manually anyway so a clear button shouldn't break anything. Especially if you make sure to pass the same value(maybe it should be empty string instead of `null`) in `onChange` as being passed if you simply clear the field with `ctrl+a` and `backspace`.

This doesn't actually trigger onChange with null. So comment is irrelevant since if we allow actual clearing it might have unexpected outcomes.

thecalcc:
> > It would be better not to affect the current UX
> 
> 
> 
> Why not @fritzSF ? If we're affecting it in a good way? :) It's not a text input, so clearing by removing invidivual characters that we have by default now is not good UX to begin with

It changes UX regardless, so would cause some amount of confusion

fritzSF:
@tomaskikutis Basically to avoid confusion in already esatablished usage scenarios, if there are any. I have to admit that I'm not completely in the loop with this specific case and it's possible implications, I can take a closer look into it. However, as a general rule, we should be cautious with changes that would affect users.

tomaskikutis:
> This doesn't actually trigger onChange with null. So comment is irrelevant since if we allow actual clearing it might have unexpected outcomes.

What's the difference between "actual clearing" and the other clearing? When you clear the input manually with backspace, what argument gets passed to `onChange`? Or `onChange` isn't called?

Also check if it's supported to pass `null` as value to datepicker. Maybe it expects a string for empty value.

tomaskikutis:
> Basically to avoid confusion in already esatablished usage scenarios

I'd agree if we were adding something significant. But I see this rather as fixing a defect that we'd want to apply everywhere. The only downside I see is in the publish pane as Konstantin mentioned where it already has a clear button that clears date and time and the same time. In that case we should pass `required="true"` to component and it wouldn't show the clearing button.

thecalcc:
> > This doesn't actually trigger onChange with null. So comment is irrelevant since if we allow actual clearing it might have unexpected outcomes.
> 
> What's the difference between "actual clearing" and the other clearing? When you clear the input manually with backspace, what argument gets passed to `onChange`? Or `onChange` isn't called?
> 
> Also check if it's supported to pass `null` as value to datepicker. Maybe it expects a string for empty value.

This is what I was saying, onChange doesn't get called when you press backspace to clear

fritzSF:
@tomaskikutis All good, I agree it is not a significant change. @thecalcc showed me the Embargo example, and although redundant the additional clear in the calandar will not block the user to clear the values they used to. If we can hide it there even better.

tomaskikutis:
What remains now is to make sure your changes don't break existing usages Konstantin. Some usages might crash if `null` is passed to `onChange`. It seems a decent workaround is to change all usages to be `required=true`?

tomaskikutis (approved):

PR #867, opened 2024-07-04
Avatar improvements and additions based on requirements for [SDESK-7311] and [SDESK-7312]

fritzSF:


dzonidoo (approved):

app-typescript/components/avatar/avatar-wrapper.tsx
```diff
@@ -14,11 +14,14 @@ interface IPropsAvatarWrapper {
     };
     children: React.ReactNode;
     'data-test-id'?: string;
-    isEmpty?: boolean;
+    emptyLight?: boolean;
```
tomaskikutis, line 17: That's not the best name. What about `theme: 'light' | 'dark'; // defaults to dark; only applies to placeholder image`

app-typescript/components/avatar/avatar-wrapper.tsx
```diff
@@ -14,11 +14,14 @@ interface IPropsAvatarWrapper {
...
     icon?: {
         name: string;
         color?: string;
     };
+    coverageStatus?: {
```
tomaskikutis, line 22: You should avoid naming that describes things outside of this repository. Maybe we intend to use it for coverage statuses, but what we're adding is a colored dot that doesn't have any limitations to usage for coverage statuses only. If we're adding a colored dot, let's call it something like that. `dot` or `statusDot` is what I'd use.

Since we want to apply it to a coverage status, and coverage status is visualized through an `icon` prop, I think it would be easier to understand the meaning if the dot was on top of the icon, not on top of the user picture. When it's on user picture it appears as it is user status like.

What happened by the way with the idea to use icon color for status? Is dot color any better? :upside_down_face: Or we still use icon color for some other meaning?

fritzSF: Makes sense, statusDot is fine. I'll change it. 

Applying it to the icon doesn't really work visually, I tried it. A good portion of the icon gets covered by the dot making it less recognisable and this way it is easier to see the dot.

The icon coloring still applies. It is used for the workflow status.

tomaskikutis: image

image


I'd do it something like that. Anyway, I'm leaving up to you to do the design it, just suggesting an idea.

app-typescript/components/avatar/avatar-wrapper.tsx
fritzSF: In this case, I would go with something that is neutral regarding the overall theme. For example, `appearance: 'subtle' | 'strong';` or something similar.

tomaskikutis: That's even worse than `emptyLight` :) what about `noAvatarPlaceholderColor: 'light' | 'dark';`. Or maybe even better - make it gray so we don't need to configurable at all :upside_down_face:

app-typescript/components/avatar/avatar-wrapper.tsx
fritzSF: It's fine, I'm generally not against it. Semanticaly this does make more sense. But we are potentially getting into issues like this:
Superdesk UI Framework 2024-07-05 at 1 52 43 PM
I believe such combinations may not be possible, but I am not sure. If this assumption is correct, we could consider the second option with a slightly smaller dot. However, if the combinations with a colored icon and a dot do exist, I will relocate it to the top right corner.

app-typescript/components/avatar/avatar-wrapper.tsx
fritzSF: How come? It's logical. I want to avoid 'light' and 'dark' because if you change the SD theme, it's no longer accurate. The 'light' option would actually be darker than the 'dark' in the dark SD theme 🙂. Hence, 'subtle' and 'strong', it holds true in both themes. As for the prop name, I don't have any preferences.

tomaskikutis: > The 'light' option would actually be darker than the 'dark' in the dark SD theme

Makes sense. So what it means in the end is whether theme color is used or inverse. What about
`noAvatarPlaceholderColor: 'theme' | 'theme-inverse'`?

app-typescript/components/avatar/avatar-wrapper.tsx
tomaskikutis: Yeah this is not ideal, but not that bad either. Since white outline exists it looks pretty well. And if we have a tooltip   describing the meaning, we're good I'd say.

PR #871, opened 2024-07-10
Close after clear

thecalcc:
SDBELGA-819

**Description:**
The popover has to be closed once the "Clear" button has been clicked because the Calendar component from primereact has to update internally and re-render the latest value in the UI. If we don't close it, once you've set a value using the quick access buttons, you have to manually open and close the popover in order for the UI value to update.


tomaskikutis:
In this case it's not obvious from a ticket why are you closing the popover. In such cases describe in the PR what is the problem you're addressing

thecalcc:
> In this case it's not obvious from a ticket why are you closing the popover. In such cases describe in the PR what is the problem you're addressing

Ok I did it

tomaskikutis (approved):

PR #872, opened 2024-07-22
Add option to disable onClose/onClick functionality

thecalcc:
SDBELGA-847

### DIFF
--- a/app-typescript/components/Layouts/Panel.tsx
+++ b/app-typescript/components/Layouts/Panel.tsx
@@ -70,6 +70,7 @@ interface IPropsPanelHeader {
     className?: string;
     onClose?(): void;
     iconButtons?: Array<React.ReactNode>;
+    disabled?: boolean;
 }
 
 class PanelHeader extends React.PureComponent<IPropsPanelHeader> {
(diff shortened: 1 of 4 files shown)


tomaskikutis:
What's the purpose of the PR? What use-case are you handling?

thecalcc:
> What's the purpose of the PR? What use-case are you handling?

Handling the case when a widget is pinned. If it is, we should disable specific onClick handlers. In this case the handles responsible for closing are the ones I've modified in this PR. Besides that the style related changes are because once pinned, when you hover on the sidebar, no close button should be shown.

tomaskikutis:
`onClose` is already optional. If you don't pass it or pass `undefined`, the close button will not be rendered. Same with hovering - if it's not rendered it won't appear in hover state. Handle this on client-core side. Is there anything else that can't be handled on client core and needs changes here?

thecalcc:
> `onClose` is already optional. If you don't pass it or pass `undefined`, the close button will not be rendered. Same with hovering - if it's not rendered it won't appear in hover state. Handle this on client-core side. Is there anything else that can't be handled on client core and needs changes here?

Removing a button if it's not needed in not the way to go about it I think. We need it to be present in the UI & make it `disabled` because of visual feedback the users need. Removing it from the UI completely would confuse the customer a lot. Besides this, for the second point - onHover is there and it'll stay, it's controlled through `activeTab`, I have to make stuff up to make it work, needs to be handled in UIF to be proper.

tomaskikutis:
> Removing it from the UI completely would confuse the customer a lot

How about not disabling closing and if user has it pinned, but still wants to close it - it would unpin and close as expected?

tomaskikutis (approved):

PR #875, opened 2024-08-20
Fix select grid

thecalcc:
SDUF-111

TODO: 
- [x] Reopening doesn't cause the picker to stop listening for key events

### DIFF
--- a/app-typescript/components/IconPicker.tsx
+++ b/app-typescript/components/IconPicker.tsx
@@ -52,25 +52,29 @@ export class IconPicker extends React.PureComponent<IProps, IState> {
         this.props.onChange(item.value);
     }
 
-    triggerTemplate = (props: any) => <Button
-        icon={this.props.value}
-        text={this.props.value}
-        onClick={(e) => { props.onClick(e); }}
-        iconOnly={true} />
+    triggerTemplate = (props: {onClick: (e: React.SyntheticEvent) => void; }) => (
+        <Button
+            icon={this.props.value}
+            text={this.props.value}
+            onClick={(e) => { props.onClick(e); }}
+            iconOnly={true}
+        />
+    )
 
-    itemTemplate = ({ item }: { item: IItem | null }) => item ?
-        (<>
+    itemTemplate = ({ item }: { item: IItem | null }) => item && (
+        <>
             <Icon name={item.value} />
             <span className="sd-text--normal sd-padding-t--1">
                 {item.label}
             </span>
-        </>) : null
+        </>
+    )
 
     render() {
         return (
             <SelectGrid
-                label={this.props.label || "Icon"}
-                filterPlaceholder={this.props.filterPlaceholder || "Search..."}
+                label={this.props.label ?? "Icon"}
+                filterPlaceholder={this.props.filterPlaceholder ?? "Search..."}
                 getItems={this.getItems}
                 onChange={this.onChange}
                 itemTemplate={this.itemTemplate}
--- a/app-typescript/components/SelectGrid.tsx
+++ b/app-typescript/components/SelectGrid.tsx
@@ -31,7 +31,7 @@ interface IState {
 
 const GRID_COLS = 3;
 const GRID_ROWS = 5;
-const PAGE_SIZE = GRID_COLS * GRID_ROWS
+const PAGE_SIZE = GRID_COLS * GRID_ROWS;
 
 export class SelectGrid extends React.PureComponent<IProps, IState> {
     htmlId = nextId();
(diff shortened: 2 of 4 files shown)


thecalcc:
Demo: 

https://github.com/user-attachments/assets/205f59fe-ec88-4a47-9f74-4d50d2ec5fe5

Note: Component updates weren't trickling down, thus the picked icon was always one step behind. I've chosen to use `this.forceUpdate()` when selecting an item to fix that. The other solution should be to pass the `value` down to the grid, but I went with `forceUpdate` as the cleaner approach.

app-typescript/components/IconPicker.tsx
```diff
@@ -52,25 +52,29 @@ export class IconPicker extends React.PureComponent<IProps, IState> {
...
+        </>
+    )
 
     render() {
         return (
             <SelectGrid
-                label={this.props.label || "Icon"}
-                filterPlaceholder={this.props.filterPlaceholder || "Search..."}
+                label={this.props.label ?? "Icon"}
+                filterPlaceholder={this.props.filterPlaceholder ?? "Search..."}
```
petrjasek, line 77: we will need those translated I guess, maybe we should rather make those props mandatory

app-typescript/components/SelectGrid.tsx
```diff
@@ -19,17 +19,21 @@ interface IProps {
...
 interface IState {
     items: Array<IItem>;
     loading: boolean;
-    isOpen: boolean;
 }
 
+const VERTICAL_OFFSET = 3;
+const HORIZONTAL_OFFSET = 1;
+const PAGE_OFFSET = 15;
+const FIRST_ROW_INDEXES = [0, 1, 2];
```
petrjasek: I wonder if it would be more readable to work with something like:
```
const GRID_COLS = 3;
const GRID_ROWS = 5;
```
or
```
const GRID_WIDTH = 3;
const GRID_HEIGHT = 5;
```
it's not clear why there would be any offsets

app-typescript/components/IconPicker.tsx
thecalcc: what if we treat those as the default fallback that's not translated, and if we need translations we pass them from the respective usage? im thinking here for all other components as well

petrjasek: we will need a mechanism to translate it within the framework probably, it wouldn't make much sense to redo it everywhere we use it. for now the fallback is ok

app-typescript/components/SelectGrid.tsx
```diff
@@ -19,17 +19,21 @@ interface IProps {
...
 }
 
 interface IState {
     items: Array<IItem>;
     loading: boolean;
-    isOpen: boolean;
 }
 
+const GRID_COLS = 3;
+const GRID_ROWS = 1;
```
petrjasek: imo there is not just 1 row, is there? in theory I would expect `PAGE_ITEMS = GRID_COLS * GRID_ROWS`

app-typescript/components/SelectGrid.tsx
```diff
@@ -19,17 +19,21 @@ interface IProps {
...
+const PAGE_ITEMS = 15;
+const FIRST_ROW_ITEMS_INDEXES = [0, 1, 2];
```
petrjasek: not sure if we need this one, you can check it via `index < GRID_COLS`

app-typescript/components/SelectGrid.tsx
thecalcc: Yep, but I renamed the offsets to `GRID_COLS` & `GRID_ROWS`. Actually they're meant to be `ROW_OFFSET` & `COL_OFFSET`. do you think those names are better?

thecalcc: confusion comes because this isn't a 2d array, it's 1d, so in theory there's no rows and cols

petrjasek: question is how many rows are actually there, so I would probably set proper value to `GRID_ROWS` so that the `PAGE_SIZE = GRID_COLS * GRID_ROWS` and where you use that `GRID_ROWS` to add/subtract `1` I would just use `1` there, that's not a magic number imo

petrjasek: and it's fine to be stored as an array, this should rather describe how it's rendered

app-typescript/components/IconPicker.tsx
thecalcc: exactly, that's why I'd leave it

petrjasek (approved):

PR #877, opened 2024-08-26
Adjust max col size

thecalcc:
SDUF-111


thecalcc:
@fritzSF I think this is a reasonable fix considering the fact that previously there was no actual max value for the columns. This was I ensure that that size is `3` and from there I can do reliable keyboard navigation, but in any case if you have a better solution I'm here

fritzSF (approved):

PR #880, opened 2024-09-17
move dev dependencies to devDependencies

petrjasek:
also don't require specific moment version, just use whatever is installed


tomaskikutis (approved):
what issues was it causing before?

petrjasek:
it's not causing issues besides installing extra packages everywhere it's used

tomaskikutis:
I see

PR #885, opened 2024-10-29
Allow custom classes for authoring header

thecalcc:
SDESK-7417


fritzSF (approved):

tomaskikutis:
If default header padding is being removed - wouldn't it work to drop that `headerStyles` prop too? We could achieve the same with a wrapper:

```
authoringHeader={(
    <div data-my-attribute="1">
        <AuthoringSection prop1="a" prop2="b"  />
    </div>
)}

PR #886, opened 2024-10-30
Fix authoring header

thecalcc:
SDESK-7341


tomaskikutis:
remove CSS classnames which are no longer used. Also do it for https://github.com/superdesk/superdesk-ui-framework/pull/885 - there too you removed usages - but not declarations

thecalcc:
> remove CSS classnames which are no longer used. Also do it for #885 - there too you removed usages - but not declarations

I've removed them in the last commit

tomaskikutis (approved):

PR #887, opened 2024-10-30
Final header fix

thecalcc:
SDESK-7417

### DIFF
--- a/app-typescript/components/Layouts/AuthoringInnerHeader.tsx
+++ b/app-typescript/components/Layouts/AuthoringInnerHeader.tsx
@@ -1,11 +1,11 @@
 import * as React from 'react';
 import classNames from 'classnames';
-import { Icon } from '../Icon';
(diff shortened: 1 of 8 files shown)


thecalcc:
Works perfectly, you can choose to only pass e.g. `top` and it would only overwrite that, without reseting the other values.

dzonidoo (approved):

app-typescript/components/Layouts/AuthoringInnerHeader.tsx
```diff
@@ -28,7 +30,12 @@ export class AuthoringInnerHeader extends React.PureComponent<IProps, IState> {
 
         return (
             <header
-                style={{padding: headerPadding}}
+                style={{
+                    paddingTop: headerPadding?.top,
```
tomaskikutis: it should be padding block start/end instead of padding top/bottom

PR #895, opened 2024-11-15
Date time picker

thecalcc:
STT-109

### DIFF
--- a/app-typescript/components/DatePicker.tsx
+++ b/app-typescript/components/DatePicker.tsx
@@ -33,6 +33,7 @@ interface IDatePickerBase extends IInputWrapper {
         }}
      */
     locale?: DatePickerLocaleSettings;
+    hideClear?: boolean;
 }
 
 interface IDatePicker extends IDatePickerBase {
@@ -160,7 +161,8 @@ export class DatePicker extends React.PureComponent<IDatePicker, IState> {
                 tabindex={this.props.tabindex}
             >
                 <Calendar
-                    footerTemplate={this.props.required !== true ? () => (
+                    className='sd-input__input'
+                    footerTemplate={this.props.hideClear !== true ? () => (
                         <div className='d-flex justify-end'>
                             <Button
                                 onClick={() => {
--- a/app-typescript/components/DatePicker.tsx
+++ b/app-typescript/components/DatePicker.tsx
@@ -33,7 +33,7 @@ interface IDatePickerBase extends IInputWrapper {
         }}
      */
     locale?: DatePickerLocaleSettings;
-    hideClear?: boolean;
+    hideClearButton?: boolean;
 }
 
 interface IDatePicker extends IDatePickerBase {
@@ -162,7 +162,7 @@ export class DatePicker extends React.PureComponent<IDatePicker, IState> {
             >
                 <Calendar
                     className='sd-input__input'
-                    footerTemplate={this.props.hideClear !== true ? () => (
+                    footerTemplate={this.props.hideClearButton !== true ? () => (
                         <div className='d-flex justify-end'>
                             <Button
                                 onClick={() => {
--- a/app-typescript/components/DatePicker.tsx
+++ b/app-typescript/components/DatePicker.tsx
@@ -34,7 +34,6 @@ interface IDatePickerBase extends IInputWrapper {
      */
     locale?: DatePickerLocaleSettings;
 
-    // Hide does not mean show
     hideClearButton?: boolean;
 }
 
--- a/app-typescript/components/DatePicker.tsx
+++ b/app-typescript/components/DatePicker.tsx
@@ -33,6 +33,8 @@ interface IDatePickerBase extends IInputWrapper {
         }}
      */
     locale?: DatePickerLocaleSettings;
+
+    // Hide does not mean show
     hideClearButton?: boolean;
 }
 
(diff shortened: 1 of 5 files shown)


app-typescript/components/DatePicker.tsx
```diff
@@ -33,6 +33,7 @@ interface IDatePickerBase extends IInputWrapper {
         }}
      */
     locale?: DatePickerLocaleSettings;
+    hideClear?: boolean;
```
tomaskikutis: Can you improve naming? [verb][verb] is a bit confusing. I'd use either `hideClearButton` or `disableClearing`

app-typescript/components/DatePicker.tsx
```diff
@@ -160,7 +161,8 @@ export class DatePicker extends React.PureComponent<IDatePicker, IState> {
                 tabindex={this.props.tabindex}
             >
                 <Calendar
-                    footerTemplate={this.props.required !== true ? () => (
+                    className='sd-input__input'
+                    footerTemplate={this.props.hideClear !== true ? () => (
```
tomaskikutis: maintain logic regarding "required"

app-typescript/components/DateTimePicker.tsx
```diff
@@ -0,0 +1,101 @@
...
+import {Spacer} from '@superdesk/common';
+import {defaultTo, padStart} from 'lodash';
+import {TimePicker} from './TimePicker';
+import {IconButton} from './IconButton';
+
+interface IProps {
+    value: Date | null;
+    label: string;
+    dateFormat: string;
+    onChange: (value?: string) => void;
```
tomaskikutis: value in `onChange` shouldn't be undefined. Maybe `string | null` ? Even if it'd be undefined, it'd be better to type `value: string | undefined`

app-typescript/components/DateTimePicker.tsx
```diff
@@ -0,0 +1,101 @@
...
+    required?: boolean;
+    width?: React.CSSProperties['width'];
+    disabled?: boolean;
+}
+
+const MIN_WIDTH = 348;
+
+export class DateTimePicker extends React.PureComponent<IProps> {
+    handleTimeChange = (time: string) => {
+        const [hours, minutes] = time.split(':').map((x) => defaultTo(parseInt(x, 10), 0)); // handle NaN value
```
tomaskikutis, line 25: FYI there's also `isNaN` function

app-typescript/components/DateTimePicker.tsx
```diff
@@ -0,0 +1,101 @@
+import * as React from 'react';
+import {DatePicker} from '../components/DatePicker';
+import {Spacer} from '@superdesk/common';
+import {defaultTo, padStart} from 'lodash';
```
tomaskikutis: can you use [padStart](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padStart) from standard library?

app-typescript/components/DateTimePicker.tsx
thecalcc: I'm aware. But in the case of using isNaN the whole expression becomes less readable in my opinion, because isNaN only checks whether the value is `NaN`, `defaultTo` does the same but also offers an immediate fallback. So in this case it is a better fit. Otherwise if you prefer a more explicit style, with isNaN it would look like `isNan(x) ? 0 : parseInt(x, 10)`.

app-typescript/components/DateTimePicker.tsx
thecalcc: Yes

app-typescript/components/DatePicker.tsx
thecalcc: This was a choice I made, because I didn't want to bloat the logic. I chose to bloat the interface instead and require users of the component to have to also pass `hideClear`. There's also a benefit to this I think - I don't couple the required logic with hiding the clear when you don't want it for other reasons.

app-typescript/components/DateTimePicker.tsx
tomaskikutis: You can leave it as you prefer, I don't mind. I just wanted to point that `isNan` function exists. Personally I slightly prefer `isNan` approach, because it doesn't use libraries and person reading the code doesn't need to understand what the library function does. I dislike lodash specifically for making the same function compatible with many data types. It's good on one side, but on another, it's making it harder to look at code and infer from usage what's going on.

app-typescript/components/DatePicker.tsx
tomaskikutis: It's a violation of a code contract between interface(props) and implementation. As a developer I say via props that field must not be empty, and you allow it to be empty - ignoring what was requested via props.


Bloat is a secondary issue compared to this. I don't see it as bloat either - being able to clear the value is a core function of any input.

app-typescript/components/DatePicker.tsx
```diff
@@ -33,6 +33,8 @@ interface IDatePickerBase extends IInputWrapper {
         }}
      */
     locale?: DatePickerLocaleSettings;
+
+    // Hide does not mean show
```
tomaskikutis: light does not mean dark ;D You'd need to phrase clearer. Or drop it completelly which would be fine here too.

tomaskikutis (approved):

PR #897, opened 2024-11-19
improve Dropdown with data-test-id

dzonidoo:


app-typescript/components/Dropdown.tsx
```diff
@@ -84,6 +84,7 @@ export const Dropdown = ({
             el.style.left = '0';
             el.style.width = '1px';
             el.style.height = '1px';
+            el.setAttribute('data-test-id', 'dropdown');
```
thecalcc: Add a data-test-id property to the component props and pass it here

tomaskikutis: only one dropdown can be open at once, so there's not much benefit into having a custom data-test-id. We'll add it if we ever find a use case. Do use a more unique data-test-id @dzonidoo . Maybe `dropdown-overlay`

tomaskikutis (approved):

PR #898, opened 2024-12-04
Fix datetime picker

thecalcc:
STT-109

### DIFF
--- a/app-typescript/components/DateTimePicker.tsx
+++ b/app-typescript/components/DateTimePicker.tsx
@@ -7,7 +7,10 @@ import {IconButton} from './IconButton';
 
 interface IProps {
     value: Date | null;
-    label?: string;
+    label: {
+        text: string;
+        hidden?: boolean;
+    };
     dateFormat: string;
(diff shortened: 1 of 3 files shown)


app-typescript/components/DateTimePicker.tsx
```diff
@@ -70,7 +70,8 @@ export class DateTimePicker extends React.PureComponent<IProps> {
                             this.handleDateChange(val?.toString());
                         }}
                         dateFormat={this.props.dateFormat}
-                        label={this.props.label}
+                        inlineLabel
```
tomaskikutis: modify the interface to make label optional and pass these props if it is not set. If a label is passed from props it should be rendered. Now you hide it unconditionally

fritzSF: Please don't forget accessibility. We should be able to define a label while choosing to hide it visually, making it accessible for screen readers. I'm not sure if this is the case here, but I'm mentioning it as a general reminder.

tomaskikutis (approved):

PR #902, opened 2025-01-21
Value type improvements

thecalcc:
STT-31


tomaskikutis (commented):
Why moment? We should try using the native date at least as input/output. In the implementation you can use moment if that helps. Even though I'd prefer date-fns because it operates on the native date.

MarkLark86:
Moment.js themselves also [recommend against using it](https://momentjs.com/docs/#/-project-status/), in favour of native date functions (or libraries that enhance it's functionality).

app-typescript/components/DateTimePicker.tsx
```diff
@@ -45,9 +45,9 @@ export class DateTimePicker extends React.PureComponent<IProps> {
         const origDate = this.props.value ?? new Date();
 
         selectedDate.setHours(origDate.getHours());
-        selectedDate.setMinutes(origDate.getMinutes());
+        selectedDate.setHours(origDate.getMinutes());
```
tomaskikutis: it should probably be `setMinutes` - you now have setHours twice

app-typescript/components/DateTimePicker.tsx
```diff
@@ -68,9 +68,9 @@ export class DateTimePicker extends React.PureComponent<IProps> {
                         preview={this.props.preview}
                         required={this.props.required}
                         hideClearButton={true}
-                        value={convertedValue}
+                        value={convertedValue ?? null}
```
tomaskikutis: if `this.props.value` is `Date | null` it doesn't make sense to have `convertedValue` variable does it?

package.json
```diff
@@ -110,5 +110,8 @@
   },
   "peerDependencies": {
     "moment": "*"
+  },
+  "volta": {
+    "node": "14.18.3"
```
tomaskikutis: can you do `v14.21.3`? I think that's the latest

tomaskikutis (approved):

PR #903, opened 2025-01-23
Stop mutating props value

thecalcc:
STT-31


app-typescript/components/DateTimePicker.tsx
```diff
@@ -26,10 +26,9 @@ const MIN_WIDTH = 348;
 export class DateTimePicker extends React.PureComponent<IProps> {
     handleTimeChange = (time: string) => {
         const [hours, minutes] = time.split(':').map((x) => defaultTo(parseInt(x, 10), 0)); // handle NaN value
-        const origDate = this.props.value ?? new Date();
+        const origDate = cloneDeep(this.props.value) ?? new Date();
```
tomaskikutis: Use clone instead of cloneDeep. Actually you don't even need lodash, to clone a date you can do
`const date2 = new Date(date1)`

tomaskikutis (approved):
fix lint

PR #905, opened 2025-01-29
Update _icon-font.scss

petrjasek:
fix sass compile error
```
SassError: Top-level selectors may not contain the parent selector "&".
    &.icon-#{$name}:before {
```


fritzSF (approved):

tomaskikutis:
don't forget to bump the version in the PR so it autoreleases after merging

PR #907, opened 2025-01-29
Fix onChange on TreeSelelect

dzonidoo:


examples/pages/components/TreeSelect.tsx
```diff
@@ -170,7 +170,7 @@ let options2 = [
 ]
 
 let fetchedArr = [];
-fetch('https://nominatim.openstreetmap.org/search/berlin?format=json&addressdetails=1&limit=20').then(response => response.json()).then(data => fetchedArr = data);
+fetch('https://nominatim.openstreetmap.org/search.php?q=berlin&exclude_place_ids=134060781&format=jsonv2').then(response => response.json()).then(data => fetchedArr = data);
```
tomaskikutis: inline the values so we don't have to rely on an external API

app-typescript/components/TreeSelect/TreeSelect.tsx
```diff
@@ -96,6 +96,12 @@ export class TreeSelect<T> extends React.Component<IProps<T>, IState<T>> {
     private popperInstance: Instance | null;
     private zIndex: number = getNextZIndex();
 
+    /*
+    * This variable is used to distinguish changes coming from outside (from props.value)
+    * from those made by the user through interaction with the component.
+    */
```
tomaskikutis, line 103: Include in the comment is that the purpose is to prevent triggering onChange when props change.

examples/pages/components/TreeSelect.tsx
```diff
@@ -169,8 +169,24 @@ let options2 = [
     },
 ]
 
-let fetchedArr = [];
-fetch('https://nominatim.openstreetmap.org/search/berlin?format=json&addressdetails=1&limit=20').then(response => response.json()).then(data => fetchedArr = data);
+let fetchedArr = [
```
tomaskikutis: since there's only item there, what's the point to have it at all? What about removing it completelly and simply using one of demo arrays we already have?

dzonidoo: initial point was to present real async mode...

tomaskikutis: async demo is using a timeout to simulate a request delay and `fetchedArr` was being accessed synchronously. It makes no difference which data source is being used if they are both accessed synchronously

tomaskikutis (approved):

PR #911, opened 2025-02-05
Togglebox performance improvements

tomaskikutis:
Do not render children if togglebox was never opened.


MarkLark86 (commented):
This works in general, but not if the toggle box includes a file input, I have found these needs to be rendered still to contain the data uploaded.
Should we have a prop maybe that forces to render children when closed?

tomaskikutis:
What's the issue with a file input? Once children mount - they will remain mounted even if togglebox is later collapsed.

MarkLark86:
> What's the issue with a file input? Once children mount - they will remain mounted even if togglebox is later collapsed.

Whoops sorry, I missed that `wasOpened` state. All good then :+1:

PR #913, opened 2025-02-13
Fix authoring-react tests

dzonidoo:
SDESK-7430

### DIFF
--- a/app-typescript/components/Select.tsx
+++ b/app-typescript/components/Select.tsx
@@ -49,7 +49,7 @@ class Select extends React.Component<ISelect> {
                         className='sd-input__select'
                         id={this.htmlId}
                         value={this.props.value}
-                        aria-describedby={this.htmlId}
+                        aria-describedby={this.htmlId + 'label'}
                         tabIndex={this.props.tabindex}
                         onChange={this.handleChange}
                         disabled={this.props.disabled || this.props.readonly}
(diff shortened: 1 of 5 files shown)


app-typescript/components/Select.tsx
```diff
@@ -49,7 +49,7 @@ class Select extends React.Component<ISelect> {
                         className='sd-input__select'
                         id={this.htmlId}
                         value={this.props.value}
-                        aria-describedby={this.htmlId}
+                        aria-describedby={this.htmlId + 'label'}
```
tomaskikutis, line 52: what is this doing?

dzonidoo: @tomaskikutis if u agree with aria-describedby={this.htmlId + 'label'} u can approve this one?

tomaskikutis: Is there a purpose for `+ 'label'`?

dzonidoo: The reason why we leave it as + 'label' is to avoid having two elements with the same Id.

tomaskikutis: I'll add that, the reason 2 IDs are needed is because label wrapper component uses one to link labels with inputs using "for" attribute - and another using "aria-described-by". The component is build in such a way that it's not possible to apply only one of them. Significant refactoring would be needed to clean this up which we won't do at this moment.

tomaskikutis (approved):

PR #915, opened 2025-02-17
Remove nav fixed height

thecalcc:
STT-129

### DIFF
--- a/app/styles/layout/_editor.scss
+++ b/app/styles/layout/_editor.scss
@@ -10,7 +10,6 @@
 .sd-editor-grid__editor-subnav {
     grid-column: 1 / 6;
     grid-row: editorSubnav;
-    height: $nav-height;
 }
 
 .sd-editor-grid__sidebar-left {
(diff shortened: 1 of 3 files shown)


thecalcc:
Removing this so that space between navbar and fields in authoring looks better. Also in my opinion if we need some default we can do `min-height` or `padding`. Or handle that in the specific usage of this navbar

Before: 
image

After:
image

tomaskikutis:
You'll be verifying usages right? Should be authoring-react only? If not, we'd need to introduce this via a prop and only do conditionally.

thecalcc:
> You'll be verifying usages right? Should be authoring-react only? If not, we'd need to introduce this via a prop and only do conditionally.

I've verified them - it's only used in authoring react, could you approve it so I can merge it to my other pr on client core

tomaskikutis (approved):

tomaskikutis:
> it's only used in authoring react,

double check if it look ok too when creating articles using authoring-react

thecalcc:
> > it's only used in authoring react,
> 
> double check if it look ok too when creating articles using authoring-react

Yep, I did check it 
image

PR #916, opened 2025-02-26
TreeMenu changes position from top to bottom when a sub category is clicked

dzonidoo:
SDESK-7255

### DIFF
--- a/app-typescript/components/TreeMenu.tsx
+++ b/app-typescript/components/TreeMenu.tsx
@@ -208,12 +208,11 @@ export class TreeMenu<T> extends React.Component<IProps<T>, IState<T>> {
                     onFirstUpdate: (state) => {
                         this.popperInstance?.setOptions({
                             placement: state.placement,
-                            modifiers: [
-                                {
-                                    name: 'flip',
-                                    enabled: false,
-                                },
-                            ],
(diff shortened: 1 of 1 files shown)


thecalcc (approved):

app-typescript/components/TreeMenu.tsx
```diff
@@ -199,6 +199,23 @@ export class TreeMenu<T> extends React.Component<IProps<T>, IState<T>> {
             if (this.openDropdownRef.current && this.dropdownRef.current) {
                 this.popperInstance = createPopper(this.openDropdownRef.current, this.dropdownRef.current, {
                     placement: 'bottom-start',
+                    modifiers: [
+                        {
+                            name: 'flip',
```
tomaskikutis, line 204: Would it work to initialize with flip modifier disabled?

dzonidoo: nope

tomaskikutis: I see. Then improve `onFirstUpdate` function to only disable flip modifier(use Array.map). At the moment you are also removing all other modifiers. We don't have any others yet, but if we added some, they would get removed after first update which is not what we want.

tomaskikutis:
fix ci

tomaskikutis (approved):

PR #917, opened 2025-03-11
Improve button component with text alignment prop

dzonidoo:


app-typescript/components/Button.tsx
```diff
@@ -35,6 +36,7 @@ export class Button extends React.PureComponent<IPropsButton> {
             'btn--icon-only': this.props.iconOnly,
             'btn--ui-dark': this.props.theme === 'dark',
             'btn--icon-only-circle': this.props.shape === 'round' && this.props.iconOnly,
+            [`sd-flex-justify-${this.props.textAlign}`]: this.props.textAlign != null,
```
tomaskikutis: do not concatenate classnames

dzonidoo: can u be more specific?

tomaskikutis: Do not use variables to generate classnames. Classnames should be static.

tomaskikutis (approved):

tomaskikutis:
merge develop - it should fix the checks failing due to that cache issue

PR #927, opened 2025-03-18
Firefox issue: Fix UI issues for Content Profile Management modal

dzonidoo:
SDESK-7529

### DIFF
--- a/app/styles/form-elements/_forms-general.scss
+++ b/app/styles/form-elements/_forms-general.scss
@@ -63,7 +63,6 @@
 
 	// For more complex inline form combinations an additional element that goes inside .form__row--flex container is provided.
 	.form__row-item {
-		flex-basis: 0;
 		flex-grow: 1;
 		margin: 0 0.6rem;
 
@@ -80,8 +79,8 @@
 		}
 	}
 
-	.sd-flex-no-grow.sd-flex-no-grow {
-		flex-basis: auto;
+	.form__row-item:not(.sd-flex-no-grow) {
+		flex-basis: 0;
 	}
 
 	.form__row-item--no-grow {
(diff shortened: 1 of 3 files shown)


app/styles/form-elements/_forms-general.scss
```diff
@@ -80,6 +80,10 @@
 		}
 	}
 
+	.sd-flex-no-grow.sd-flex-no-grow {
```
tomaskikutis: Instead of overwriting the value - rather do not apply it in the first place if not needed:

`.form__row-item:not(.sd-flex-no-grow) { flex-basis: 0 }`

dzonidoo: but we need both scenarios

tomaskikutis: Can you explain in more detail? To me it seems that the issue is caused by addition of `flex-basis:0`. So it's enough not to add it and the issue is solved.

fritzSF: I really don't think there is a need for a detailed explaneation of such basic stuff, but here you go:` flex-basis: 0;` should be the default; it makes the field items equal in size, which is what we want 90% percent of the time. `.sd-flex-no-grow ` is an override of the default behaviour—we don't want the the field to grow, and we want it to have `flex-basis: auto;`, exactly as in the casementioned in the ticket.

tomaskikutis: I understand well the issue and the solution. Unless you see a scenario where the snippet I posted wouldn't work - let's discuss it. If not, do adjust the PR to use it.

tomaskikutis (approved):

PR #929, opened 2025-03-30
use same style tooltip for buttons as everywhere else

tomaskikutis:
SDESK-7489

### DIFF
--- a/app-typescript/components/Button.tsx
+++ b/app-typescript/components/Button.tsx
@@ -1,7 +1,8 @@
 import * as React from 'react';
 import classNames from 'classnames';
-import { Icon } from './Icon';
-import { Spinner } from './Spinner';
+import {Icon} from './Icon';
+import {Spinner} from './Spinner';
+import {WithTooltip} from './Tooltip';
 
 interface IPropsButton {
     text: string;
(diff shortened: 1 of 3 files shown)


app-typescript/components/Tooltip.tsx
```diff
@@ -24,14 +25,18 @@ function flowToPlacement(flow: IProps['flow']): Placement | undefined {
...
+export class WithTooltip extends React.PureComponent<IProps> {
     private id: string;
     private instance: Instance | null;
 
     constructor(props: IProps) {
         super(props);
 
-        this.id = 'tooltip-' + Math.random().toString().slice(2);
+        // prepend a letter so ID doesn't start with a number
+        this.id = 'a' + Math.random().toString().slice(2);
```
thecalcc: and this can't instead be `${Math.random().toString().slice(2)}`?

app-typescript/components/Button.tsx
```diff
@@ -41,23 +42,40 @@ export class Button extends React.PureComponent<IPropsButton> {
...
+                    <WithTooltip text={this.props.tooltip}>
+                        {({attributes}) => children({attributes})}
+                    </WithTooltip>
+                )
+                : ({children}) => (
+                    <>
+                        {children({attributes: {}})}
+                    </>
+                );
+
```
thecalcc: may be better to move this out of here, I know it's specifically made for the button but maybe add it below the button class

thecalcc (approved):

app-typescript/components/Tooltip.tsx
tomaskikutis: https://css-tricks.com/ids-cannot-start-with-a-number/

Your `Math.random().toString()` would generate a string, but when you use numbers as HTML attributes they become strings too. So ID not staring with a number means number in the string.

app-typescript/components/Button.tsx
```diff
@@ -75,7 +64,23 @@ export class Button extends React.PureComponent<IPropsButton> {
...
+            <WithTooltip text={tooltipText}>
+                {({attributes}) => children({attributes})}
+            </WithTooltip>
+        )
+        : ({children}) => (
+            <>
+                {children({attributes: {}})}
+            </>
+        );
+}
```
thecalcc: is there some specific reason you made it a function and not a component with props? Just curious

tomaskikutis: Lack of imagination I guess. I needed a wrapper component, so made a function that returns one. I would be curious to see how would a component with props work here if at all.

thecalcc: Not sure I get it, but it would work like any other HOC:

```
interface ITooltipWrapperProps {
    tooltipText: string | null | undefined;
    children: React.ComponentProps<typeof WithTooltip>['children'];
}

class TooltipWrapper extends React.Component<ITooltipWrapperProps> {
    render() {
        const {tooltipText, children} = this.props;

        return (tooltipText ?? '').length > 0
            ? (
                <WithTooltip text={tooltipText}>
                    {({attributes}) => children({attributes})}
                </WithTooltip>
            )
            : (
                <>
                    {children({attributes: {}})}
                </>
            );
    }
}
```

thecalcc (approved):

PR #930, opened 2025-04-03
TreeSelect improvements

dzonidoo:
SDBELGA-945

### DIFF
--- a/app-typescript/components/Form/InputWrapper.tsx
+++ b/app-typescript/components/Form/InputWrapper.tsx
@@ -13,6 +13,7 @@ export interface IInputCommon {
     labelHidden?: boolean;
     tabindex?: number;
     fullWidth?: boolean;
+    autoWidth?: boolean;
     boxedStyle?: boolean;
     boxedLable?: boolean;
 }
@@ -54,6 +55,7 @@ export class InputWrapper extends React.Component<IPropsBase, IState> {
             [`sd-input--${this.props.size}`]: this.props.size || this.props.size !== undefined,
             'sd-input--boxed-style': this.props.boxedStyle,
             'sd-input--boxed-label': this.props.boxedLable,
+            'sd-input--auto-width': this.props.autoWidth,
         });
 
         const labelClasses = classNames('sd-input__label', {
--- a/app-typescript/components/Form/InputWrapper.tsx
+++ b/app-typescript/components/Form/InputWrapper.tsx
@@ -12,8 +12,13 @@ export interface IInputCommon {
     inlineLabel?: boolean;
     labelHidden?: boolean;
     tabindex?: number;
+
+    /**
+     * Defaults to `true`
+     * If set to false, it will be as wide as content
+     */
     fullWidth?: boolean;
-    autoWidth?: boolean;
+
     boxedStyle?: boolean;
     boxedLable?: boolean;
 }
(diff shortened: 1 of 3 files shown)


tomaskikutis:
Can you make it work without extra props being necessary?

We already have a prop `inputWidth?: '100%'`. If it is not passed, the component should be as wide as contents.

dzonidoo:
The inputWidth & fullWidth props no longer make sense because we've improved the grid, and now components have width: 100% by default. it should be removed.
and use the autoWidth if u want to component be as wide as contents...

dzonidoo:
this is the PR when i added that one input width
https://github.com/superdesk/superdesk-ui-framework/pull/809

tomaskikutis:
> The inputWidth & fullWidth props no longer make sense

Can they be removed then?

> use the autoWidth if u want to component be as wide as contents

Can you adjust the code so width based on contents is default and `100%` is optional prop?

dzonidoo:
> Can they be removed then?
- yep sure... ill just spent more time in investigation to be sure... there is a lot of component with inputWrapper

> Can you adjust the code so width based on contents is default and `100%` is optional prop?
- hmmm nope... That could cause mass in other components too and on cc on planning as well etc..

tomaskikutis:
I see. `100%` width by default is fine if that's how it works already for all other inputs.

The problem is that it collapses to zero in some cases.

I did some testing and I got it to be either 100% or squashed to contents - with no conditional logic required. See if it's possible to make it work similarly without passing the prop.

```typescript
import React from 'react';

export function App(props) {
    const testBox = <div style={{background: 'red', border: '1px solid blue'}}>aaa</div>;
    return (
        <div style={{background: 'white', width: '100vw', height: '100vh'}}>
            {testBox}

            <table>
                <tbody>
                    <tr>
                        <td>one</td>
                        <td>two</td>
                    </tr>
                    <tr>
                        <td>{testBox}</td>
                        <td>four</td>
                    </tr>
                </tbody>
            </table>
        </div>
    );
}
``` 

[rendered result](https://github.com/user-attachments/assets/0898ee6c-078a-4c4c-996c-b1118bbde66b)

fritzSF:
The component has a quite complex layout, and changing it without breaking anything would require experimentation. A prompt is, in my opinion, completely legitimate; it collapses to zero only in some edge cases. It is easier to change the surrounding layout or use the prompt than to modify the component itself. I say we leave it as is and consider this approach for future improvements.

app-typescript/components/Form/InputWrapper.tsx
```diff
@@ -13,6 +13,7 @@ export interface IInputCommon {
     labelHidden?: boolean;
     tabindex?: number;
     fullWidth?: boolean;
+    autoWidth?: boolean;
```
tomaskikutis: I see there's `fullWidth` prop above. How many width configurations are possible now? Only "full/auto" or there are others?

fritzSF:
We only need "full/auto", with _full_ being the default.

tomaskikutis:
Are there more options currently implemented that we don't need anymore?

fritzSF:
No, not for now

tomaskikutis:
I've looked into it. The issue originates from a label wrapper component. We should fix it, maybe simplify the component at the same time. @fritzSF do you remember if it contains anything special besides rendering a label below or at the side?

tomaskikutis:
Here's an example where it doesn't render well in either case if label length is not known. Which is the case here since it's coming from a vocabulary.

image

image

fritzSF:
I investigated yesterday different cases, I think I have a fix for that.

PR #939, opened 2025-05-27
implement custom time popover

thecalcc:
UI screenshots for a configuration with AM/PM and seconds:
image
image

### DIFF
--- a/app-typescript/components/TimePicker.tsx
+++ b/app-typescript/components/TimePicker.tsx
@@ -6,7 +6,7 @@ import {TimePickerPopover} from './TimePickerPopover';
 import {PopupPositioner} from './ShowPopup';
 
 interface IProps extends IInputWrapper {
-    value?: string; // ISO8601 time string(e.g. 16:55) or null if there's no value
+    value: string | null; // ISO8601 time string(e.g. 16:55) or null if there's no value
     onChange(valueNext: string): void;
     allowSeconds?: boolean;
     headerTemplate?: React.ReactNode;
@@ -83,8 +83,7 @@ export class TimePicker extends React.PureComponent<IProps, IState> {
                         cursor: 'pointer',
                     }}
                     ref={this.timeInputRef}
-                    key={this.props.value}
-                    value={this.props.value}
+                    value={this.props.value ?? undefined}
                     type="time"
                     onClick={(e) => {
                         // don't show default popup
--- a/app-typescript/components/TimePicker.tsx
+++ b/app-typescript/components/TimePicker.tsx
@@ -1,56 +1,86 @@
-import * as React from 'react';
-import nextId from 'react-id-generator';
-import {InputWrapper} from './Form';
-import {IInputWrapper} from './Form/InputWrapper';
+import * as React from "react";
+import nextId from "react-id-generator";
+import { InputWrapper } from "./Form";
+import { IInputWrapper } from "./Form/InputWrapper";
+import { WithPopover } from "./WithPopover";
+import { TimePickerPopover } from "./TimePickerPopover";
 
 interface IProps extends IInputWrapper {
-    value: string; // will output time as ISO8601 time string(e.g. 16:55) or an empty string if there's no value
-    onChange(valueNext: string): void;
-    allowSeconds?: boolean;
-    'data-test-id'?: string;
+  value?: string; // will output time as ISO8601 time string(e.g. 16:55) or an empty string if there's no value
+  onChange(valueNext: string): void;
+  allowSeconds?: boolean;
+  headerTemplate?: React.ReactNode;
+  footerTemplate?: React.ReactNode;
+  "data-test-id"?: string;
 }
 
 export class TimePicker extends React.PureComponent<IProps> {
-    private htmlId = nextId();
+  private htmlId = nextId();
 
-    render() {
-        if (this.props.preview) {
-            return (
-                <div>
-                    <span>{this.props.value}</span>
-                </div>
-            );
-        }
-
-        return (
-            <InputWrapper
-                label={this.props.label}
-                error={this.props.error}
-                invalid={this.props.error != null}
-                required={this.props.required}
-                disabled={this.props.disabled}
-                info={this.props.info}
-                inlineLabel={this.props.inlineLabel}
-                labelHidden={this.props.labelHidden}
-                htmlId={this.htmlId}
-                tabindex={this.props.tabindex}
-                inputWrapper={this.props.inputWrapper}
-            >
-                <input
-                    value={this.props.value}
-                    type="time"
-                    className="sd-input__input"
-                    id={this.htmlId}
-                    aria-labelledby={this.htmlId + 'label'}
-                    step={this.props.allowSeconds ? 1 : undefined}
-                    required={this.props.required}
-                    disabled={this.props.disabled}
-                    onChange={(event) => {
-                        this.props.onChange(event.target.value);
-                    }}
-                    data-test-id={this.props['data-test-id']}
-                />
-            </InputWrapper>
-        );
+  render() {
+    if (this.props.preview) {
+      return (
+        <div>
+          <span>{this.props.value}</span>
+        </div>
+      );
     }
+
+    return (
+      <InputWrapper
+        label={this.props.label}
+        error={this.props.error}
+        invalid={this.props.error != null}
+        required={this.props.required}
+        disabled={this.props.disabled}
+        info={this.props.info}
+        inlineLabel={this.props.inlineLabel}
+        labelHidden={this.props.labelHidden}
+        htmlId={this.htmlId}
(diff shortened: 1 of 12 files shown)


app-typescript/components/TimePickerPopover.tsx
```diff
@@ -0,0 +1,316 @@
...
+    super(props);
+
+    this.activeTime = [React.createRef(), React.createRef(), React.createRef()];
+
+    this.parseUnitOfTime = this.parseUnitOfTime.bind(this);
+    this.handleChange = this.handleChange.bind(this);
+
+    const hour = new Date().toLocaleTimeString([], { hour: "numeric" });
+
+    this.is12HourFormat = hour.includes("AM") || hour.includes("PM");
```
tomaskikutis: Accept `locale?: {code?: string, timeFormat?: 12 | 24}`

Priority would be timeFormat, locale code from props, or current version where browser locale is used

app-typescript/components/TimePickerPopover.tsx
```diff
@@ -0,0 +1,316 @@
...
+      </span>
+    );
+  }
+}
+
+export class TimePickerPopover extends React.PureComponent<IProps> {
+  is12HourFormat: boolean;
+
+  // hour, minutes, seconds
+  activeTime: Array<React.RefObject<TimeValueHolder>>;
```
tomaskikutis: Declare explicitly whether members are public/private. Use a name that indicates that refs are stored. Maybe `inputRefs`?

app-typescript/utils/time.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+export type ITimeUnit = 'hours' | 'minutes' | 'seconds';
+
+export function getOptionsForTimeUnit(
+    timeUnit: ITimeUnit,
+    is12HourFormat?: boolean,
+    disabledOptions?: {[key: string]: Array<number>},
+): Array<string> {
+    const format12HourArr = range(1, 13);
+    format12HourArr.unshift(format12HourArr.pop() as number);
```
tomaskikutis: how about you create the final range that you want above so you don't need another step to adjust it?

package-lock.json
```diff
@@ -663,7 +663,7 @@
     "acorn-dynamic-import": {
       "version": "2.0.2",
       "resolved": "https://registry.npmjs.org/acorn-dynamic-import/-/acorn-dynamic-import-2.0.2.tgz",
-      "integrity": "sha1-x1K9IQvvZ5UBtsbLf8hPj0cVjMQ=",
+      "integrity": "sha512-GKp5tQ8h0KMPWIYGRHHXI1s5tUpZixZ3IHF2jAu42wSCf6In/G873s6/y4DdKdhWvzhu1T6mE1JgvnhAKqyYYQ==",
```
tomaskikutis: package.json hasn't changed - you probably don't need package-lock changes either?

app-typescript/components/TimePicker.tsx
```diff
@@ -2,11 +2,15 @@ import * as React from 'react';
 import nextId from 'react-id-generator';
 import {InputWrapper} from './Form';
 import {IInputWrapper} from './Form/InputWrapper';
+import {WithPopover} from './WithPopover';
+import {TimePickerPopover} from './TimePickerPopover';
 
 interface IProps extends IInputWrapper {
-    value: string; // will output time as ISO8601 time string(e.g. 16:55) or an empty string if there's no value
+    value?: string; // will output time as ISO8601 time string(e.g. 16:55) or an empty string if there's no value
```
tomaskikutis: > or an empty string if there's no value

If we support an empty string, which is fine, let's keep `value` mandatory.

app-typescript/components/TimePicker.tsx
```diff
@@ -34,22 +38,48 @@ export class TimePicker extends React.PureComponent<IProps> {
...
-                    aria-labelledby={this.htmlId + 'label'}
-                    step={this.props.allowSeconds ? 1 : undefined}
-                    required={this.props.required}
-                    disabled={this.props.disabled}
-                    onChange={(event) => {
-                        this.props.onChange(event.target.value);
-                    }}
-                    data-test-id={this.props['data-test-id']}
-                />
+                <WithPopover
```
tomaskikutis: add data-test-id for popover too

app/styles/_time.scss
```diff
@@ -0,0 +1,28 @@
+.time-unit-highlight {
+    background-color: $sd-colour-interactive;
```
tomaskikutis: only use CSS variables

app/styles/_time.scss
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    display: flex;
+    justify-content: center;
+    width: 32px;
+    height: 32px;
+    border-radius: $border-radius__base--x-small;
+    font-weight: normal;
+    margin-inline-start: 4px;
+    margin-inline-end: 4px;
```
tomaskikutis: use design tokens for sizes e.g. `--base-increment`, `--gap-1`

app-typescript/utils/time.tsx
```diff
@@ -0,0 +1,32 @@
+import {range, padStart} from 'lodash';
```
tomaskikutis: use `String.prototype.padStart`

app-typescript/components/TimePickerPopover.tsx
```diff
@@ -0,0 +1,262 @@
...
+
+        this.parseUnitOfTime = this.parseUnitOfTime.bind(this);
+        this.handleChange = this.handleChange.bind(this);
+
+        const hour = new Date().toLocaleTimeString([], {hour: 'numeric'});
+
+        this.is12HourFormat = hour.includes('AM') || hour.includes('PM');
+    }
+
+    parseUnitOfTime(unit: ITimeUnit) {
```
tomaskikutis: it would be easier to understand if it was a pure function at the top level

app-typescript/components/TimePickerPopover.tsx
```diff
@@ -0,0 +1,262 @@
...
+            marginTop: 8,
+        };
+
+        return (
+            <div className="sd-shadow--z2 radius-md" onBlur={this.props.closePopup}>
+                <Spacer
+                    v
+                    gap="0"
+                    style={{
+                        padding: 8,
```
tomaskikutis: use CSS variables for sizes where applicable

app-typescript/components/TimePickerPopover.tsx
```diff
@@ -0,0 +1,262 @@
...
+                        width: 200,
+                        backgroundColor: 'white',
```
tomaskikutis: use CSS variables for colors so theming works ok

app-typescript/utils/time.tsx
thecalcc: here I move last number `12` to first place. I copied the code from another component since I think it's good. The other option is to manually insert `12` at index `0` with `arrayInsertAtIndex` from `@superdesk/common` and change the array creation to `const format12HourArr = range(1, 12);`. Let me know which one you prefer.

app/styles/_time.scss
thecalcc: This is a `scss` file, why would I do that?

tomaskikutis: we'll be dropping sass variables in favor of CSS ones that are standard according to spec and also it's very useful that we can use CSS variables from JS which we can't do with sass ones

app/styles/_time.scss
```diff
@@ -0,0 +1,28 @@
...
+
+    display: flex;
+    justify-content: center;
+    width: 32px;
+    height: 32px;
+    border-radius: var(--b-radius--x-small);
+    font-weight: normal;
+    margin-inline-start: var(--gap--x-small);
+    margin-inline-end: var(--gap--x-small);
+    transition: $sd-transition__menu-item;
```
tomaskikutis: use a CSS variable here

app-typescript/components/TimePickerPopover.tsx
```diff
@@ -0,0 +1,290 @@
...
+                        <>
+                            {this.props.headerTemplate}
+                            <ContentDivider border type="solid" orientation="horizontal" margin="none" />
+                        </>
+                    )}
+                    <Spacer h gap="4" noWrap justifyContent="center" alignItems="start">
+                        <Spacer v gap="4" style={styleForColumnOfUnit} alignItems="center" noWrap>
+                            {getOptionsForTimeUnit('hours', this.is12HourFormat).map((hour) => {
+                                /**
+                                 * Hour value is always in 24-hour format, so we need to adjust it
```
tomaskikutis: you're already passing `this.is12HourFormat` to `getOptionsForTimeUnit` - that function could return the final version, couldn't it?

tomaskikutis (approved):

tomaskikutis (approved):

PR #941, opened 2025-05-30
Fix inconsistent color and time clearing

thecalcc:
SDESK-7632

### DIFF
--- a/app/styles/form-elements/_checkbox.scss
+++ b/app/styles/form-elements/_checkbox.scss
@@ -675,8 +675,8 @@ $checkButtonBorderRadius: $border-radius__base--small;
         & ~ label {
             opacity: 1;
             color: $white;
-            background-color: var(--sd-colour-interactive--active);
-            border-color: var(--sd-colour-interactive--active);
+            background-color: var(--sd-colour-interactive);
+            border-color: var(--sd-colour-interactive);
             border-top-color: var(--sd-colour-interactive--darken-20);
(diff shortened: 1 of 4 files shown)


app/styles/form-elements/_checkbox.scss
```diff
@@ -672,29 +696,31 @@ $checkButtonBorderRadius: $border-radius__base--small;
...
 
     .sd-check-button__input:checked {
-        & ~ label {
+        &~label {
             opacity: 1;
             color: $white;
-            background-color: var(--sd-colour-interactive--active);
-            border-color: var(--sd-colour-interactive--active);
+            background-color: var(--sd-colour-interactive);
+            border-color: var(--sd-colour-interactive);
```
thecalcc, line 679: @tomaskikutis this is the only real change in this file. I could not avoid formatting the whole file.

tomaskikutis: Can't you configure your editor not to touch certain files? .scss files are ignored in .prettierignore

Undo the formatting changes. It's not a sustainable approach to work like this. Either you adjust your editor or we can talk about re-formattig everything with prettier if reconfiguring your editor doesn't work.

thecalcc: Yes, now I configured my editor to not format only scss files on save. The problem I see with this approach is that everyone will have to do it, instead a better one would be to reformat all files with the prettier configuration you setup. I though we had that already

tomaskikutis (approved):

PR #945, opened 2025-06-17
TimePicker and DatePicker improvements

fritzSF:


package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "superdesk-ui-framework",
-    "version": "4.0.56",
```
thecalcc: version here now is `.57`, that should also change in `package-lock.json` to match it.

app-typescript/components/TimePicker.tsx
```diff
@@ -93,7 +100,22 @@ export class TimePicker extends React.PureComponent<IProps, IState> {
...
                     }}
-                    className="sd-input__input"
+                    onKeyDown={(e) => {
+                        if (e.key === ' ') {
+                            // don't show default popup
+                            e.preventDefault();
+
+                            this.setState({
+                                popupOpen: true,
+                            });
```
thecalcc: So if user presses any key, he'll open the popover? should we make it so it's like a toggle on enter? If user presses Enter and popover is open, we close it, and if it's not open we do open it. 

Then for any other keys pressed we simply do `e.preventDefault()` without opening our popover?

app-typescript/components/TimePicker.tsx
```diff
@@ -104,6 +126,18 @@ export class TimePicker extends React.PureComponent<IProps, IState> {
...
                 />
+                <div className="sd-input__icon">
+                    <Icon name="time" />
+                    <div id="clear-time-picker">
+                        <IconButton
+                            icon="remove-sign"
+                            size="small"
+                            ariaValue="Clear"
+                            toolTipFlow="left"
+                            onClick={this.props.onClear ? this.props.onClear : this.handleTimeClear}
```
thecalcc: what's the difference between `onClear` and `handleTimeClear`? I guess if we want to cleanup time only or date and time in the component we should abstract that higher up, and only use `onClear` here

thecalcc:
@fritzSF let me know if you need help with code changes

app-typescript/components/TimePicker.tsx
```diff
@@ -4,13 +4,16 @@ import {InputWrapper} from './Form';
...
+import {IconButton} from './IconButton';
 
 interface IProps extends IInputWrapper {
     value: string | null; // ISO8601 time string(e.g. 16:55) or null if there's no value
-    onChange(valueNext: string): void;
+    onChange(valueNext: string | null): void;
     allowSeconds?: boolean;
     headerTemplate?: React.ReactNode;
     footerTemplate?: React.ReactNode;
+    onClear?: () => void;
```
tomaskikutis: Why would we need `onClear` in the interface? If we have a clear button, it should execute `this.props.onChange(null)` instead.

app-typescript/components/TimePickerPopover.tsx
```diff
@@ -133,25 +133,35 @@ export class TimePickerPopover extends React.PureComponent<IProps> {
...
+            scrollBehavior: 'smooth',
         };
 
         return (
-            <div className="sd-shadow--z2 radius-md" onBlur={this.props.closePopup}>
+            <div
+                className="sd-shadow--z2 radius-md"
+                onBlur={this.props.closePopup}
+                onKeyDown={(e) => {
+                    if (e.key === 'Enter') {
```
tomaskikutis: It would be good to stick to default keybindings, which is `space` to open, `ESC` to close.

app-typescript/components/TimePicker.tsx
```diff
@@ -104,6 +126,18 @@ export class TimePicker extends React.PureComponent<IProps, IState> {
                     }}
                     data-test-id={this.props['data-test-id']}
                 />
+                <div className="sd-input__icon">
+                    <Icon name="time" />
+                    <div id="clear-time-picker">
```
tomaskikutis: use a classname. IDs must be unique in HTML document, and with multiple time pickers on the page - it wouldn't be.

app/styles/form-elements/_inputs.scss
```diff
@@ -783,23 +783,88 @@
...
+                .icon-time {
+                    opacity: 0;
+                }
+                #clear-time-picker {
+                    opacity: 1;
+                }
+            }
+        }
+    }
+    #time-picker { // only for the DateTimePicker component. Disable the clearing of the time input only.
```
tomaskikutis: can you separate time picker styles into its own file?

examples/pages/components/TimePicker.tsx
```diff
@@ -19,13 +21,28 @@ class TimePickerExample extends React.PureComponent<{}, {time: string | null}> {
...
+                }
+                footerTemplate={<div>Footer</div>}
                 value={this.state.time}
                 onChange={(time) => {
                     this.setState({time});
                 }}
                 allowSeconds
                 label="This is Label"
                 info="This is info"
+                inputWrapper={{kind: 'custom', component: ({input}) => (
```
tomaskikutis: adjust the styles so it looks well when a custom wrapper is used

image

tomaskikutis: feel free to undo this `inputWrapper` demo, I just added it so you can reproduce it breaking

app-typescript/components/TimePicker.tsx
fritzSF: _Space_ opens the popover if the input field is focused. _Enter_ accepts the changes and closes the popup. It would be helpful to include _Escape_ to discard the changes and close the popup, as the native one in Chrome does.

app-typescript/components/TimePicker.tsx
fritzSF: Yes, this was intended to enable the clearing of the time alone in the DateTimePicker. I wasn't able to achieve it though, so I will remove it.

app-typescript/components/TimePickerPopover.tsx
fritzSF: `space` to open, `Enter` to close it and accept the changes, `ESC` to close it and discard the changes, like the native one.

app-typescript/components/TimePicker.tsx
thecalcc: Got it, I'd suggest then we use `e.keyCode === 32` (`32` can be extracted into some constant like `const SPACE_KEY = 32` for better readability) or even better `e.code === 'Space'` since it's harder to read the `' '` in my opinion

tomaskikutis: [keyCode](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode) is deprecated. Using `e.key === ' '` is how it should be done. One possible alternative is to use a constant instead of inlined string with a space.

app-typescript/components/TimePicker.tsx
fritzSF: done.

app-typescript/components/TimePicker.tsx
fritzSF: done.

app-typescript/components/TimePicker.tsx
fritzSF: done.

app/styles/form-elements/_inputs.scss
fritzSF: Done, it should now work fine with a custom wrapper.

app-typescript/components/TimePicker.tsx
tomaskikutis: We usually don't use the "done" comments. The person reviewing checks the changes after review and either resolves comments on github or replies with more information. I'll check your changes :)

tomaskikutis:
Is this animation intended?

https://github.com/user-attachments/assets/6e0b55fe-c04b-4916-85fe-e8f0d520ac1d

tomaskikutis:
probably not :upside_down_face:

https://github.com/user-attachments/assets/7deeb5f0-b740-4304-a22d-a45c75f4d3b3

app/styles/form-elements/_inputs.scss
tomaskikutis: There was another comment about input wrapper, this one is about splitting time picker styles into its own file.

fritzSF:
The scroll animation is intended

> Is this animation intended?
> screen-recording-2025-06-19_12.36.24.mp4

The scroll animation is intended. Smoth scrolling will be needed to keep the selected item in scroll.

fritzSF:
> probably not 🙃
> screen-recording-2025-06-19_12.38.38.mp4

This beheaviour is not. The pop-up should not reload on each click. I didn't behave like that, I'll have to check.

fritzSF:
> probably not 🙃
> screen-recording-2025-06-19_12.38.38.mp4

https://github.com/user-attachments/assets/d9d47904-3c35-48f7-80a3-0eca21d6ed73

It seems that this happens only with the custom wrapper.

thecalcc:
@tomaskikutis cc @fritzSF the issue with the scroll on each change is mine, don't worry about it, I'll see how to fix it

thecalcc:
> @tomaskikutis cc @fritzSF the issue with the scroll on each change is mine, don't worry about it, I'll see how to fix it

I think I did some re-rendering after value changes, so the new value is reflected in the popup

tomaskikutis:
I've checked the issue of animation happening on every click. It was due to usage of custom wrapper template. I've fixed it.

app-typescript/components/DateTimePicker.tsx
```diff
@@ -173,7 +173,8 @@ export class DateTimePicker extends React.PureComponent<IProps> {
                             data-test-id="date-input"
                         />
                     </div>
-                    <div style={{flexGrow: 1}}>
+                    <div style={{flexGrow: 0, lineHeight: '3.2rem', color: 'var(--color-text-muted)'}}>@</div>
```
tomaskikutis: `line-height` is only used for centering here which stops working if input field height changes. Coupling can be avoided and code stability improved if centering was handled by flex parent. Not sure why is `align-items` is set to `end` at the moment, but setting it to `center` handles vertical centering without introducing coupling.

Is `flexGrow` doing anything? It renders the same without it.

tomaskikutis (commented):
Could you create a standalone stylesheet files for datepicker, timepicker and datetime picker, co-locate them next to react components and ensure that each stylesheets only targets elements defined in their respective component?

app-typescript/components/DateTimePicker.tsx
fritzSF: The input height will not change, as it has the same hardcoded value. We should replace it with a variable at some point, but this is fine for now. 

Spacer settings where not introduced in this PR or by me. I would say that it is set to `end` because of the possibility of individual labals for date and time; `center` would misalign the '@' and the button at the end.

`flexGrow: 1` on the inputs is necessary. `flexGrow: 0` on the '@' is indeed not and can be removed.

fritzSF:
The CSS splitting and relocation can be handled as a separate task since it doesn't impact the core functionality. I've created a ticket for myself and will address it after my vacation: https://sofab.atlassian.net/browse/SDESK-7710. @thecalcc has also a ticket prepared for the keyboard navigation improvements and can proceed with it when ready.

tomaskikutis:
I've merged current improvements @fritzSF. For keyboard navigation we don't have time next week weeks, we may do that sometime later.

PR #950, opened 2025-08-04
Refactor badge and button styles for improved theming and consistency. 

fritzSF:
- Adjust form element heights to use new design tokens.
- Add new button type 'tertiary' and enhance examples for better clarity


thecalcc:
Which commits have your actual changes @fritzSF? Let me know and I'll cherry pick them to a new branch based on the latest develop so it's easier to read

app/styles/editor/_editor-themes.scss
```diff
@@ -91,14 +103,19 @@
...
 
 		--sd-editor-colour__mark-company:      var(--syrah-soil-300);
 
 		--sd-editor-colour__soft-limit:     var(--syrah-soil-200);
 		--sd-editor-colour__hard-limit:     var(--morocco-red-400);
+
+		--sd-editor-colour__mark-company:      var(--syrah-soil-300);
+
+		--sd-editor-colour__soft-limit:     var(--syrah-soil-200);
+		--sd-editor-colour__hard-limit:     var(--morocco-red-400);
```
tomaskikutis: this duplicates lines above

app/styles/editor/_editor-themes.scss
```diff
@@ -543,3 +571,106 @@ $body-font-size: (
 		}
 	}
 }
+
+
+// TEMP ================================
```
tomaskikutis: I undestand this is demo code? Can you remove it?

app/styles/form-elements/_inputs.scss
```diff
@@ -739,6 +739,43 @@
     color: $red;
 }
 
+input[type="color"] {
```
tomaskikutis: can you target a class instead so we don't lose the capability to have multiple implementations running at the same time?

app/styles/form-elements/_inputs.scss
```diff
@@ -739,6 +739,43 @@
...
+    transition: all ease 0.2s;
+    &:hover {
+        border-color: var(--color-input-border-hover);
+        cursor: pointer;
+    }
+    &:focus-visible {
+        outline: 2px solid var(--color-interactive-shadow-focus);
+    }
+}
+input[type="color" i]::-webkit-color-swatch-wrapper {
```
tomaskikutis: what does `i` mean?

app-typescript/components/Button.tsx
```diff
@@ -31,9 +40,12 @@ export class Button extends React.PureComponent<IPropsButton> {
     render() {
         let classes = classNames('btn', {
             'btn--expanded': this.props.expand,
+            'btn--hollow': this.props.type === 'tertiary' || this.props.style === 'hollow',
```
tomaskikutis: Keep props independent. Settting one shouldn't have an effect on another.

package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "superdesk-ui-framework",
-    "version": "4.0.72",
+    "version": "4.0.73",
```
tomaskikutis, line 3: don't edit this file manually, run `npm version patch` instead so package-lock file gets updated too.

app/styles/form-elements/_inputs.scss
fritzSF: It stands for case-insensitive matching, so `[type="color"]`, `[type="Color"]`, etc.

app/styles/form-elements/_inputs.scss
fritzSF: I don't see any situation where I would want it to have default styling.

tomaskikutis: Regardless of predictions, code should stick to standards. `ui-framework` exports components - components have own styles that must not affect anything outside components. Global styles violate this.

fritzSF: I could agree to a certain degree, but I disagree that this should be a general rule. There are no real standards for some elements, including this one; each browser handles it differently—just look at the pop-up that opens.

It would be much less error-prone to have global styles for certain elements. Someone needs to review all PRs and check if the proper class is applied to avoid issues like this:

Superdesk - Settings - Metadata 2025-08-06 at 4 53 35 PM

fritzSF: > ui-framework exports components - components have own styles that must not affect anything outside components.

This is not entirely true. While it does this, it also defines many global styles in the CSS.

tomaskikutis: I agree that default styles might not look as we like - and that's expected. Probably no element looks as we'd like without styling. But that styling musn't be global in order not to conflict with parallel implementations. We might only have one at the moment, but that's just a quality rule not to impose a restriction of having only a single implementation by doing styling via classes.

> This is not entirely true. While it does this, it also defines many global styles in the CSS.

That's one of thing we should be improving going forward. An easy first step is stop adding new globals.

fritzSF: I think there are valid reasons to consider global styling here that outweigh those concerns, but this probably isn't the best place to get into all that. Happy to discuss it elsewhere if you want.

tomaskikutis (approved):

app-typescript/components/Button.tsx
tomaskikutis: I've dropped tertiary style to keep props independent and also interface clean. Naming isn't the best in that component - in reality - `type` means color. Tertiary isn't bringing a new color. Also doesn't suport `filled` style and looks identical to `secondary/hollow` - which we can already use.

fritzSF: Aaah! I somehow didn't send the reply on your first comment! Here it is now, although late:
This is an intermediate state, I don't want to affect the existing style options. I will optimise this once we get rid of the Hollow option, which we currently have all over accross SD and Planning. The goal is three button levels 'Primary', 'Secondary' and 'Tertiary'.

PR #951, opened 2025-08-12
Add date locale support datetime picker

thecalcc:
SDBELGA-966

### DIFF
--- a/app-typescript/components/DateTimePicker.tsx
+++ b/app-typescript/components/DateTimePicker.tsx
@@ -10,11 +10,8 @@ import {format} from 'date-fns';
 import {assertNever} from '../helpers';
 import {Button} from './Button';
 
-interface IPropsValueDate extends IInputWrapper {
-    valueType: 'date';
-    value: Date | null;
+interface IBaseProps extends IInputWrapper {
     dateFormat: string;
-    onChange: (value: Date | null) => void;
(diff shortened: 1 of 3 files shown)


app-typescript/components/DateTimePicker.tsx
```diff
@@ -24,28 +21,25 @@ interface IPropsValueDate extends IInputWrapper {
     'data-test-id'?: string;
     timeHeaderTemplate?: React.ReactNode;
     timeFooterTemplate?: React.ReactNode;
+    dateLocale?: DatePicker['props']['locale'];
```
petrjasek: why `dateLocale` and not just `locale`

thecalcc: because this component is combining date and time pickers, and I'm not sure how we might want to handle time locale in the future. might need an additional prop like `timeLocale` or something else. 

And also type of `DatePicker['props']['locale']` in one case includes only the `code` in the other case there's no time related information

petrjasek: I wonder if we would ever use different locale for date and time at the same time, that sounds pretty unlikely

thecalcc: you're right we probably won't ;-)

petrjasek (approved):

PR #952, opened 2025-08-12
UI issue in dark mode with location and attached files dropzone

dzonidoo:
STT-1253


package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "superdesk-ui-framework",
-    "version": "4.0.73",
```
thecalcc: update the version

thecalcc (approved):

PR #957, opened 2025-08-28
Custom search for tree select

tomaskikutis:
SDBELGA-967


thecalcc (approved):
do we use short names like `arr` for `array`?

PR #960, opened 2025-09-03
Added support for small tag size in Tag component

fritzSF:
- Added size prop to the Tag component.
- Updated examples to demonstrate new tag sizes.
- Tag component SCSS updated.
- Removed `overflow: hidden;` from `.sd-list-item__row` in `_list-item.scss`

### DIFF
--- a/app/styles/components/_list-item.scss
+++ b/app/styles/components/_list-item.scss
@@ -269,7 +269,7 @@ $sd-ListItem-column-border: var(--sd-colour-line--x-light);
     flex-direction: row;
     justify-content: flex-start;
     align-items: center;
-    overflow: hidden;
+    // overflow: hidden;
     gap: var(--gap--small);
 
     &:first-child {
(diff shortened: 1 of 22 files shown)


app/styles/components/_list-item.scss
```diff
@@ -269,7 +269,7 @@ $sd-ListItem-column-border: var(--sd-colour-line--x-light);
     flex-direction: row;
     justify-content: flex-start;
     align-items: center;
-    overflow: hidden;
+    // overflow: hidden;
```
tomaskikutis: should that instead be removed?

tomaskikutis (approved):

app/styles/components/_list-item.scss
fritzSF: Yes, I'll remove it.

tomaskikutis (approved):

PR #962, opened 2025-09-17
WrapperButton component & doc pages updates

fritzSF:


app-typescript/components/WrapperButton.tsx
```diff
@@ -0,0 +1,68 @@
...
+        return (
+            <TooltipWrapper tooltipText={this.props.tooltip}>
+                {({attributes}) => (
+                    <button
+                        {...attributes}
+                        id={this.props.id}
+                        className={classes}
+                        tabIndex={0}
+                        disabled={this.props.disabled}
+                        onClick={this.props.disabled ? () => false : (event) => this.props.onClick(event)}
```
tomaskikutis: No need to check for disabled here, if button's `disabled` property is true, `onClick` won't get called at all.

app-typescript/components/WrapperButton.tsx
```diff
@@ -0,0 +1,68 @@
+import * as React from 'react';
+import classNames from 'classnames';
+import {WithTooltip} from './Tooltip';
+
+interface IPropsWrapperButton {
+    text?: string;
```
tomaskikutis: It will work to pass text as children e.g. `<WrapperButton>abc</WrapperButton>`. Having `text` as separate prop isn't providing any additional function.

app-typescript/components/WrapperButton.tsx
```diff
@@ -0,0 +1,68 @@
...
+    id?: string;
+    children?: React.ReactNode;
+    disabled?: boolean;
+    ariaLabel?: string;
+    radius?: 'x-small' | 'small' | 'medium' | 'full';
+    className?: string;
+    'data-test-id'?: string;
+}
+
+export class WrapperButton extends React.PureComponent<IPropsWrapperButton> {
```
tomaskikutis: `HeadlessButton` would be a better name.

app-typescript/components/WrapperButton.tsx
```diff
@@ -0,0 +1,68 @@
...
+                'radius-full': this.props.radius === 'full',
+            },
+            this.props.className,
+        );
+
+        return (
+            <TooltipWrapper tooltipText={this.props.tooltip}>
+                {({attributes}) => (
+                    <button
+                        {...attributes}
```
tomaskikutis: attributes should be spread last so they take precedence

app-typescript/components/WrapperButton.tsx
```diff
@@ -0,0 +1,68 @@
...
+    className?: string;
+    'data-test-id'?: string;
+}
+
+export class WrapperButton extends React.PureComponent<IPropsWrapperButton> {
+    render() {
+        let classes = classNames(
+            'wrapper-button',
+            {
+                'wrapper-button--disabled': this.props.disabled,
```
tomaskikutis: this class will never be used since CSS is targeting `disabled` attribute already.

tomaskikutis (approved):

PR #963, opened 2025-09-17
Support jsx as tooltip content

tomaskikutis:
SDESK-7766

Removed tippy.js and instead made a tooltip using `WithPopover` component. Support is kept for the old interface.

image

### DIFF
--- a/app-typescript/components/Tooltip.tsx
+++ b/app-typescript/components/Tooltip.tsx
@@ -1,15 +1,9 @@
 import * as React from 'react';
-import nextId from 'react-id-generator';
-import tippy, {Instance, Placement} from 'tippy.js';
 import {assertNever} from '../helpers';
+import {IPropsTooltipV2, TooltipV2} from './TooltipV2';
+import {Placement} from 'popper.js';
 
-interface IProps {
-    text: string | undefined | null;
-    flow?: 'top' | 'left' | 'right' | 'down'; // defaults to 'top'
-    children(options: {attributes: {[name: string]: string}}): React.ReactNode;
-}
-
-function flowToPlacement(flow: IProps['flow']): Placement | undefined {
+function flowToPlacement(flow: IPropsLegacy['flow']): Placement | undefined {
     switch (flow) {
         case undefined:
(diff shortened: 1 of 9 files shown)


thecalcc (approved):

app-typescript/components/TooltipV2.tsx
```diff
@@ -0,0 +1,75 @@
...
+import {Placement} from '@popperjs/core';
+import {WithPopover} from './WithPopover';
+
+export interface IPropsTooltipV2 {
+    content: string | React.ComponentType;
+    placement?: Placement;
+
+    /**
+     * If unsure - use ReactNode.
+     * Function is for advanced use cases where it's needed to wrapper span.
```
thecalcc: I don't understand the comment

app-typescript/components/Tooltip.tsx
```diff
@@ -26,83 +20,29 @@ function flowToPlacement(flow: IProps['flow']): Placement | undefined {
...
-                    <div {...attributes} style={{display: 'inline-flex'}}>
-                        {this.props.children}
-                    </div>
-                )}
-            </WithTooltip>
-        );
+        if (this.props.content != null) {
+            return <TooltipV2 {...this.props} />;
+        } else {
+            // backwards compatibility for legacy props
```
thecalcc, line 36: how many usages do we have across the projects? may be better to go ahead and replace them all until we have it under control

app-typescript/components/TooltipV2.tsx
tomaskikutis: how about now?

app-typescript/components/Tooltip.tsx
tomaskikutis: Around 20 maybe. We can use the new interface for new usages so old usages won't grow. It'll be easier to replace it when it can be done separatelly for each project and each branch.

app-typescript/components/TooltipV2.tsx
thecalcc: looks good

PR #967, opened 2025-10-02
Improve CSS for menu

thecalcc:
STT-1338

Improve nested menu items to support long text items. Previously this caused menu items to get positioned in a way where they were blocking the rest of the menu items. 

done by @fritzSF

### DIFF
--- a/app/styles/primereact/_pr-menu.scss
+++ b/app/styles/primereact/_pr-menu.scss
@@ -1,4 +1,5 @@
 .p-tieredmenu {
+    min-width: 200px;
     &, .p-submenu-list {
         padding-block-start: $sd-base-increment;
         padding-block-end: $sd-base-increment;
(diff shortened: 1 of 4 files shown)


fritzSF (approved):

app/styles/primereact/_pr-menu.scss
```diff
@@ -1,15 +1,35 @@
...
+        width: 100%;
+        max-width: 36ch;
+    }
+    .p-menuitem-text {
+        overflow: hidden;
+        text-overflow: ellipsis;
+        white-space: nowrap;
+    }
+    .p-submenu-list {
+        .p-menuitem-text {
```
tomaskikutis, line 20: Consider dropping nesting when there's only one nested item. prefer `.p-submenu-list .p-menuitem-text {` instead.

I wouldn't put ellipsis on a menu item either. Following good UX practices menu options aren't supposed to be long. And if it can't be avoided and in some cases they do need to be long - then it doesn't make sense to hide part of it using ellipsis. Multi line menu item would be a better option.

tomaskikutis (commented):
Could you add a "before" screenshot showing the issue?

app/styles/primereact/_pr-menu.scss
fritzSF: before
Menu options are not supposed to be long, I completelly agree. In this specific case though, we are passing trough the headline of an article, so it could potentially get very long. It's a valid point regarding ellipsis, but for now we can go with it.

PR #968, opened 2025-10-06
implement on blur handler for time picker

tomaskikutis:
STT-1342

Required to sort and deduplicate times in planning view.

image

### DIFF
--- a/app-typescript/components/TimePicker.tsx
+++ b/app-typescript/components/TimePicker.tsx
@@ -76,18 +76,12 @@ export class TimePicker extends React.PureComponent<IProps, IState> {
                     <PopupPositioner
                         getReferenceElement={() => this.timeInputRef.current as HTMLElement}
                         shouldClose={(event) => {
-                            const clickedInsideInput
-                                = this.timeInputRef
-                                    ?.current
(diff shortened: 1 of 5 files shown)


app-typescript/components/TimePicker.tsx
```diff
@@ -25,16 +29,24 @@ interface IState {
...
         super(props);
 
         this.timeInputRef = React.createRef();
+        this.popupPositionerRef = React.createRef();
         this.state = {
             popupOpen: false,
         };
     }
 
+    componentDidUpdate(_prevProps: Readonly<IProps>, prevState: Readonly<IState>): void {
```
thecalcc, line 44: no way for this to use a close handler of the popover? maybe get called on unmount in the child component?

thecalcc (approved):
probably needs a new version

app-typescript/components/TimePicker.tsx
tomaskikutis: It could, but it'd need to be in 2 places - in normal `onClose` and in when handling keyboard events. I thought I'd put it here so it's in one place.

PR #969, opened 2025-10-09
add navigators

taylorli2000:
SDCP-996

add navigators needed in https://github.com/superdesk/superdesk-cp/pull/474


app-typescript/components/DatePicker.tsx
```diff
@@ -13,7 +13,13 @@ import {getMonthNames, getWeekdayNames} from '@sourcefabric/common';
 import {localization} from '../localization';
 import {assertNever} from '../helpers';
 
-interface IDatePickerBase extends IInputWrapper {
+interface NavigatorsDateRange {
+    showNavigators?: boolean;
```
tomaskikutis: Don't expose these as option - show navigators unconditionally.

app/styles/components/_datepicker.scss
```diff
@@ -0,0 +1,15 @@
+.datepicker-calendar-panel {
+  div {
+    div {
+      div:nth-child(3) {
```
tomaskikutis: Use cleaner selectors. No need to count children or use nesting. Primereact exposes everything needed as classnames

image

tomaskikutis (commented):
Adjust styles to maintain existing design:

image

app-typescript/components/DatePicker.tsx
```diff
@@ -176,6 +179,15 @@ export class DatePicker extends React.PureComponent<IDatePicker, IState> {
 
         const showClearButton = this.props.required === true ? false : this.props.hideClearButton !== true;
 
+        const navigatorsProps = {
```
tomaskikutis: Iinline these props where they're used. It's only used once thus there is no use to move it into a constant.

tomaskikutis (commented):
image

There's an issue with caret alignment when short month names are used. That's because native select doesn't support having width same as currently selected value.

I've released `@superdesk/primereact@5.0.2-16` where [I've implemented passing custom select components for month/year](https://github.com/superdesk/primereact/commit/23de78c886589cb8da85120c1aa407bf8eec2741).

Use it pass a select component implementation that doesn't have the issue with caret alignment. It should work to configure `TreeSelect` for this - it supports custom templates.

tomaskikutis:
image

Also use this font icon instead of adding a new image file. I understand it probably wasn't possible with native select - but will be possible when using `TreeSelect`

taylorli2000:
> image
> There's an issue with caret alignment when short month names are used. That's because native select doesn't support having width same as currently selected value.
> 
> I've released `@superdesk/primereact@5.0.2-16` where [I've implemented passing custom select components for month/year](https://github.com/superdesk/primereact/commit/23de78c886589cb8da85120c1aa407bf8eec2741).
> 
> Use it pass a select component implementation that doesn't have the issue with caret alignment. It should work to configure `TreeSelect` for this - it supports custom templates.

I've updated prime react to pass in the tree selects as custom selects. I added a clearable prop to hide the remove button since required and readonly don't fit here. carets are now aligned as well and styled accordingly

app-typescript/components/DatePicker.tsx
```diff
@@ -334,3 +334,41 @@ export class DatePickerISO extends React.PureComponent<IDatePickerISO> {
...
+
+const YearNavigator = ({
+    value,
+    options,
+    onChange,
+}: React.ComponentProps<NonNullable<CalendarProps['yearNavigatorComponent']>>) => (
+    <TreeSelect<string>
+        kind="synchronous"
+        value={[value]}
+        getOptions={() => options.map((o) => ({...o, value: o.label}))}
```
tomaskikutis: Do not use naming shorhands - use `option` instead of `o`.

Another issue is that your return value doesn't match the interface. Due to use of spread operator typescript isn't flagging it. That's a good reason to avoid spreads.

tomaskikutis (commented):
https://github.com/user-attachments/assets/9c6f174d-2c7d-457d-8988-5cfdd18b96d3

- [x] fix date picker closing after selecting a month/year
- [x] improve CSS code quality - only add styles for classnames that you define - use `valueTemplate` prop of tree select to pass a custom template.

tomaskikutis:
Also update styles to match the screenshot https://github.com/superdesk/superdesk-ui-framework/pull/969#pullrequestreview-3319761730 Currently font colors/weight and spacing are not matching. Use CSS existing CSS variables to find similar colors for text.

app-typescript/components/TreeSelect/TreeSelect.tsx
```diff
@@ -273,6 +274,11 @@ export class TreeSelect<T> extends React.Component<IProps<T>, IState<T>> {
                 }
             });
 
+            this.dropdownRef.current?.addEventListener('mousedown', (e: MouseEvent) => {
```
tomaskikutis: Is this one to prevent date picker closing after selecting a month/year? If so the code should go to date picker - not tree select.

app/styles/primereact/_pr-datepicker.scss
```diff
@@ -9,13 +9,80 @@
...
+            button {
+                &:focus-visible {
+                    background-color: hsla(214, 13%, 55%, 0.2);
+                    box-shadow: $icn-button-focus-box-shadow;
+                }
+            }
+        }
+
+        .sd-datepicker__navigator-value {
+            z-index: -1;
```
tomaskikutis: z index isn't needed here is it?

app/styles/primereact/_pr-datepicker.scss
```diff
@@ -9,13 +9,80 @@
         line-height: 1;
         margin: 0 auto;
         text-align: center;
+        display: flex;
+        flex-direction: column;
+        align-items: center;
+        gap: $sd-base-increment * 0.2;
+
+        .sd-datepicker__navigator-wrapper {
+            * {
```
tomaskikutis: avoid star operator, list all elements you are targetting explicitly

app-typescript/components/DatePicker.tsx
tomaskikutis: I see you still haven't addressed types not matching. Drop spread operator and you'll see the typescript issue

PR #970, opened 2025-10-13
Update color variables across styles

fritzSF:
-Changed `--color-text-light` to `--color-text-muted`, and `--color-text-lighter` to `--color-text-subdued` for theme agnostic naming and adjusted various components to reflect these changes
- Introduced new Form layout guideline page
- Fixed Spinner component issues in Firefox
- Updated doc pages
- Added Colors playground page


app/styles/_helpers.scss
```diff
@@ -1970,3 +2120,130 @@ h6 {
...
+    color: var(--color-info-text);
+}
+
+
+// ------------------------------------------------------------------------
+.left-panel {
+    container-type: inline-size;
+    container-name: left-panel;
+}
+.new-form-layout {
```
thecalcc, line 2207: would it make sense to rename this to just `form-layout` or it would cause conflicts with old styles?

examples/pages/components/utilities/TextUtilities.tsx
```diff
@@ -335,25 +335,39 @@ class TextUtilitiesDoc extends React.Component {
...
-                                    <td>{'color: var(--color-text--inverse) !important;'}</td>
-                                    <td style={{backgroundColor: 'var(--color-text-lighter)'}}>
+                                    <td>{'color: var(--color-text-inverse) !important;'}</td>
+                                    <td style={{backgroundColor: 'var(--color-text-subdued)'}}>
                                         <span className="text-color-inverse">Aa</span>
                                     </td>
                                 </tr>
+                                <tr>
+                                    <td>text-color-on-dark</td>
+                                    <td>{'color: var(--color-text-ondark) !important;'}</td>
```
thecalcc, line 359: could the `!important` be removed and we implement it in another way where this color takes priority over whatever other color we have?

examples/pages/components/utilities/TextUtilities.tsx
```diff
@@ -335,25 +335,39 @@ class TextUtilitiesDoc extends React.Component {
...
+                                    <td style={{backgroundColor: 'var(--color-surface-dark)'}}>
+                                        <span className="text-color-on-dark">Aa</span>
+                                    </td>
+                                </tr>
+                                <tr>
+                                    <td>text-color-on-light</td>
+                                    <td>{'color: var(--color-text-onlight) !important;'}</td>
```
thecalcc, line 366: same question here about `!important`

thecalcc (approved):
looks good, couple small things

app/styles/_helpers.scss
fritzSF: This is temporary, for testing purposes at the moment. I will rename it once I have a proper solution.

examples/pages/components/utilities/TextUtilities.tsx
fritzSF: It could, in theory, but not at the moment. The logic is that the utility classes should override default classes, if they are there. Hence the `!important`. We could implement `@layer` rules for utility classes at some point. If applied corrextly, it would change the default priority. I would need to restructure the CSS and test it.

PR #971, opened 2025-10-14
[STT-1364] - Allow popover to be shown on hover

thecalcc:
STT-1364

Add an option for popover to be shown when user hovers the target element.


tomaskikutis:
See also my changes where I implemented show on hover function [for tooltip](https://github.com/superdesk/superdesk-ui-framework/blob/develop/app-typescript/components/TooltipV2.tsx). If you implement support for hovering in the positioner itself it would be good to eventually update the tooltip to use the same.

Main issue with your implementation is that popover closes when you try to move the mouse inside it - which means no buttons,inputs or anything interactive is supported inside. I didn't implement it either in my tooltip component, but purpose of a tooltip is more limited.

thecalcc:
> See also my changes where I implemented show on hover function [for tooltip](https://github.com/superdesk/superdesk-ui-framework/blob/develop/app-typescript/components/TooltipV2.tsx). If you implement support for hovering in the positioner itself it would be good to eventually update the tooltip to use the same.
> 
> Main issue with your implementation is that popover closes when you try to move the mouse inside it - which means no buttons,inputs or anything interactive is supported inside. I didn't implement it either in my tooltip component, but purpose of a tooltip is more limited.

I agree on the main issue, I thought of it and decided we can just have it more limited for now. It shouldn't be that hard to add it in the future - we do an extended timeout for closing, so user has time to hover over tooltip, and if they do, we pause the close timer until they don't go outside of the content.

app-typescript/components/_Positioner.tsx
```diff
@@ -100,28 +104,41 @@ class PopperWrapper extends React.Component<IPropsPopperWrapper> {
...
+interface IPropsPositioner extends IBasePositioner {
+    type: 'click';
+}
+
+interface IPropsPositionerOnHover extends IBasePositioner {
+    type: 'hover';
+}
+
+type IProps = IPropsPositioner | IPropsPositionerOnHover;
+
```
eos87, line 131: I think this can be simplified. Unless there is another reason to have two separate interfaces.
```suggestion
interface IProps extends IBasePositioner {
    type: 'click' | 'hover';
}

```

eos87 (approved):
Looks good, just left a minor suggestion.

app-typescript/components/_Positioner.tsx
thecalcc: rather a good practice, leave room open when different props apply to different types. We showcase that it can be type narrowed

eos87: but do you think `IPropsPositioner` and `IPropsPositionerOnHover` will be extended further from what you have there?

PR #972, opened 2025-10-14
Timepicker ux improvements

tomaskikutis:
STT-1381


thecalcc (commented):
could you post a before/after screenshot?

tomaskikutis:
Only scrollbars would be visbile in the screenshot. Here's a video that shows keyboard input and popover change where value is only set when all inputs are filled.

https://github.com/user-attachments/assets/9e4ffd1f-9b72-42ae-909b-0529d4a80766

app-typescript/components/TimePicker/TimePickerPopover.tsx
```diff
@@ -0,0 +1,284 @@
...
+        seconds: seconds ?? secondsDefault,
+        period: hours == null ? null : isAm(parseInt(hours, 10)) ? 'am' : 'pm',
+    };
+}
+
+const is12HourFormat: boolean = (() => {
+    const hour = new Date().toLocaleTimeString([]);
+
+    return hour.includes('AM') || hour.includes('PM');
+})();
```
eos87: it would be good to have this util functions covered with a few tests.

app-typescript/components/TimePicker/index.tsx
```diff
@@ -120,18 +120,12 @@ export class TimePicker extends React.PureComponent<IProps, IState> {
                         ref={this.timeInputRef}
                         value={this.props.value ?? ''}
                         type="time"
-                        onClick={(e) => {
-                            // don't show default popup
-                            e.preventDefault();
-
```
eos87, line 126: wouldn't this trigger the browser's built-in picker?

app-typescript/components/TimePicker/index.tsx
```diff
@@ -120,18 +120,12 @@ export class TimePicker extends React.PureComponent<IProps, IState> {
...
+                        onClick={() => {
                             this.setState({
                                 popupOpen: !this.state.popupOpen,
                             });
                         }}
                         onKeyDown={(event) => {
-                            // don't show default popup
-                            event.preventDefault();
-
```
eos87, line 134: same here?

app-typescript/components/TimePicker/index.tsx
tomaskikutis: No. There's a special icon in the native time picker that triggers the default browser UI and we have that icon hidden with CSS so it can't be clicked.

image

app-typescript/components/TimePicker/index.tsx
tomaskikutis: yep

app-typescript/components/TimePicker/index.tsx
eos87: gotcha!

eos87 (approved):

PR #978, opened 2025-11-03
Fix am pm switching

thecalcc:
SDCP-1018


thecalcc:
merge without squashing

petrjasek (approved):

PR #981, opened 2025-11-04
SearchBar & OverflowStack

fritzSF:
**Changes on the SerchBar component**
- Added external triggering
- Added Debounced automatic search
- various CSS improvements
- Doc page created
**New OverflowStack component**
A flexible container that stacks any child elements. It can display a fixed number of items and move the remainder into a popover. Optionally, it can automatically adjust which items are visible based on available space, hiding the excess in the popover.
Features:
- 3 gap variations
- Ovelap mode
- Option to show all or just hidden items in the popover
- Dynamic Overflow (based on available space)
- Indicator style (hidden item count or 3 dot icon)
- Custom Popover Content
- Custom Click Handler (not sure if this is needed at all)
- Data-Driven API (Different rendering of items in the stack vs popover)


thecalcc:
hi @fritzSF could you in the future make smaller PRs that are focused on one component/feature area? It's a bit hard to review a bunch of things with a lot of line changes. Reviews will be of better quality if we do that I think

Edit: if the feature itself is single and too big, we can split it in parts. I didn't cover that case with my comment

app-typescript/components/SearchBar.tsx
```diff
@@ -4,10 +4,13 @@ import {Icon} from './Icon';
...
 interface IProps {
     value?: string;
-    type?: 'expanded' | 'collapsed' | 'boxed';
+    type?: 'expanded' | 'collapsed';
     placeholder: string;
     focused?: boolean;
     boxed?: boolean;
+    hideSearchButton?: boolean; // Hide the internal search button (useful when triggering search externally)
+    searchOnType?: boolean; // Enable automatic search while typing (debounced)
+    searchDelay?: number; // Delay in milliseconds for searchOnType (default: 300)
```
thecalcc: we can use discriminated union here to make the interface cleaner and the component easier to use I think. Maybe we could even separate the core of the component from these two types of search and make two. Let me know if you can implement it with Cursor, I can also implement it

app-typescript/components/SearchBar.tsx
```diff
@@ -20,31 +23,122 @@ interface IState {
...
         super(props);
         this.state = {
-            inputValue: this.props.value ? this.props.value : '',
-            focused: this.props.focused ? this.props.focused : false,
-            type: this.props.type ? this.props.type : 'expanded',
-            boxed: this.props.boxed ? this.props.boxed : false,
+            inputValue: this.props.value || '',
+            focused: this.props.focused || false,
+            type: this.props.type || 'expanded',
+            boxed: this.props.boxed || false,
```
thecalcc: we can keep the existing more explicit code style here. Also the difference between `||` and `??` is that `||` will take any falsy value - i.e. `""` and `undefined` counts as falsy in the case of `||`, which might not be the intended behaviour. same goes for expressions using `? :`

thecalcc:
@eos87 could you take a look also. I need to spend more time thinking this through but I don't think we need the `simple` version to have `renderPopoverItem`. If user wants to have it different they could use `data` type. I also need to think more on naming

thecalcc:
@eos87 i made some changes to the SearchBar also. Generally now input state is handled internally, but I'm not sure yet if the following approach is better:

We keep input value storage outside the component and then get rid of all the exposed functions like `focus`, `search`, etc. Since if we store the value outside, we can also execute the search from anywhere. And for the debounce search, that's fine since it's just a wrapper on top of the submit function. Thoughts? in my opinion, this approach would keep the component lighter while also allowing for an external search trigger

eos87 (approved):
It looks good to me

PR #983, opened 2025-11-11
Add `LoadMoreIndicator` component and docs

eos87:
### Purpose
Introduces a new `LoadMoreIndicator` React component for use in infinite scroll lists. Updates exports and documentation to include usage examples, props, accessibility notes, and TypeScript config for examples.

### What has changed
- New functional component for showing loading state during infinite loading
- Includes configurable loader size, progress counter, and custom text
- Accessible with proper ARIA attributes (role, aria-live, aria-busy)
- Complete documentation with live examples and usage patterns
- Added TypeScript config for examples/components folder to fix JSX warnings

### How to test
- Run the project 
- Go to http://127.0.0.1:9100/#/components/load-more-indicator
- Observe the loader in action

#### This is how it looks
image

[STT-1453]

[STT-1453]: https://sofab.atlassian.net/browse/STT-1453?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app-typescript/components/Lists/LoadMoreIndicator.tsx
+++ b/app-typescript/components/Lists/LoadMoreIndicator.tsx
@@ -31,7 +31,7 @@ interface IProps {
      * Size of the loader
      * @default 'small'
      */
-    loaderSize?: 'small' | 'medium' | 'large';
+    size?: 'small' | 'medium' | 'large';
 
     /**
      * Custom className for the container
@@ -66,7 +66,7 @@ export const LoadMoreIndicator: React.FC<IProps> = ({
     totalCount,
     loadingText = 'Loading more...',
     showProgress = true,
-    loaderSize = 'small',
+    size = 'small',
     className,
     'data-test-id': testId = 'load-more-indicator',
 }) => {
@@ -76,9 +76,9 @@ export const LoadMoreIndicator: React.FC<IProps> = ({
 
     let loaderClassName = 'sd-loader';
 
-    if (loaderSize === 'small') {
+    if (size === 'small') {
         loaderClassName = 'sd-loader sd-loader--small';
-    } else if (loaderSize === 'large') {
+    } else if (size === 'large') {
         loaderClassName = 'sd-loader sd-loader--large';
     }
 
--- a/app-typescript/components/Lists/LoadMoreIndicator.tsx
+++ b/app-typescript/components/Lists/LoadMoreIndicator.tsx
@@ -96,7 +96,10 @@ export const LoadMoreIndicator: React.FC<IProps> = ({
                 <span className="sd-text__normal">
                     {loadingText}
                     {showProgress && currentCount != null && totalCount != null && (
-                        <> ({currentCount} of {totalCount})</>
+                        <>
+                            {' '}
+                            ({currentCount} of {totalCount})
+                        </>
                     )}
                 </span>
             </div>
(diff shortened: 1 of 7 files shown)


app-typescript/components/Lists/LoadMoreIndicator.tsx
```diff
@@ -0,0 +1,108 @@
...
+     */
+    loaderSize?: 'small' | 'medium' | 'large';
+
+    /**
+     * Custom className for the container
+     */
+    className?: string;
+
+    /**
+     * Test ID for testing purposes
```
thecalcc: not sure if we need so much comments on the props, since they're pretty standard

app-typescript/components/Lists/LoadMoreIndicator.tsx
```diff
@@ -0,0 +1,108 @@
...
+     * Whether to show the progress counter (e.g., "25 of 100")
+     * @default true
+     */
+    showProgress?: boolean;
+
+    /**
+     * Size of the loader
+     * @default 'small'
+     */
+    loaderSize?: 'small' | 'medium' | 'large';
```
thecalcc: we normally use `size`, i think it makes sense given we've got a very definitive component name

app-typescript/components/Lists/LoadMoreIndicator.tsx
```diff
@@ -0,0 +1,108 @@
...
+interface IProps {
+    /**
+     * Whether the loading indicator should be visible
+     */
+    loading: boolean;
+
+    /**
+     * Current number of loaded items
+     */
+    currentCount: number;
```
thecalcc, line 12: would be good to combine this prop and the next one together with `showProgress` so we have something like 

```ts
progressConfig: {
  showProgress: true;
  currentCount: number;
  totalCount: number;
}
```

just a bit cleaner imo

eos87: I'd lean towards keeping them separate. If you think about it `currentCount` and `totalCount` aren't really config but dynamic data that change with every render so I rather keep them separate.

app-typescript/components/Lists/LoadMoreIndicator.tsx
eos87: done!

app-typescript/components/Lists/LoadMoreIndicator.tsx
eos87: removed it for test id. I'll keep the others. My intention would be in the near future to create a POC with StoryBook and these would really help with documentation.

app-typescript/components/Lists/LoadMoreIndicator.tsx
thecalcc: that's fine, it's still good to have them together since we won't show them if `showProgress` is `false`. this is just about making the interface cleaner and easier to work with.

So if you decide to configure the progress with `progressConfig` or whatever you want to name it, all the 3 of them are mandatory. if we do that option you can even remove `showProgress `. 

You could name it `progress` i guess also. not trying to nitpick, these small things just add up over time and you end up with interfaces like we have them for text inputs

thecalcc (approved):

PR #990, opened 2025-11-27
Example

thecalcc:
### Screenshots

[ticket-number]

### DIFF
--- a/app-typescript/components/TreeSelect/TreeSelect.tsx
+++ b/app-typescript/components/TreeSelect/TreeSelect.tsx
@@ -5,7 +5,7 @@ import nextId from 'react-id-generator';
 import _debounce from 'lodash/debounce';
 import {InputWrapper} from '../Form';
 import {createPopper, Instance} from '@popperjs/core';
-import {isEqual} from 'lodash';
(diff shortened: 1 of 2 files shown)


app-typescript/components/TreeSelect/TreeSelect.tsx
```diff
@@ -147,6 +147,10 @@ export class TreeSelect<T> extends React.Component<IProps<T>, IState<T>> {
         this.changesFromOutside = false;
     }
 
+    shouldComponentUpdate(nextProps: Readonly<IProps<T>>, nextState: Readonly<IState<T>>): boolean {
+        return !isEqual(this.props, nextProps) || !isEqual(this.state, nextState);
```
petrjasek, line 151: would rather check `isEqual(this.props.value, nextProps.value)` otherwise you're testing it as before

thecalcc: that's just an example, i won't push this one, this is the real one - https://github.com/superdesk/superdesk-ui-framework/pull/991

we can also check not just value but all non function value props i think?

PR #996, opened 2025-12-04
[SDESK-7725] - Introduce new component Textbox with Copy button

thecalcc:
### Purpose

Create a new text box with a copy button for the product API client secret and ID.

### Screenshot
image

[SDESK-7725]

[SDESK-7725]: https://sofab.atlassian.net/browse/SDESK-7725?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


thecalcc:
@fritzSF could you help me out with the right tokens for the colours, I'm not sure which ones you want to use for border and background

eos87:
@thecalcc I added a screenshot just for doc purposes :)

app-typescript/components/CopyableTextBox.tsx
eos87, line 1: two minor things I would suggest: 
- let's use more meaningful names for props and state like `ICopyableTextBoxProps` or similar
- we should use function based components if possible. React is moving away from class based ones and this one seems pretty simple to tackle :)

thecalcc: totally agree with both comments, I just go on autopilot creating this component

thecalcc: i'll make those changes

fritzSF (approved):
I would just remove the different size options from the doc page. We would never use a large input for this purpose

eos87 (approved):

PR #997, opened 2025-12-04
[SDESK-7830] Upgrade Node to `v22` and drop support for `v14`

thecalcc:
### Purpose

Upgrade UI framework to node 22

### What has changed

- Upgraded webpack to v4, so we can start also using node 22.
- Migrated from `node-sass` to `sass` since it's deprecated.
- Migrated from `extract-text-webpack-plugin` to mini css extract plugin since it's deprecated.
- Updated React to latest v16 version so `https://github.com/superdesk/react-resizable-panels` dependency resolution works.
- Removed `eslint-loader` package since it's not needed.
- Updated other packages to higher versions so they work with webpack v4.

[SDESK-7830]

[SDESK-7830]: https://sofab.atlassian.net/browse/SDESK-7830?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


thecalcc:
@eos87 i updated GH actions and now i'm not 100% sure where the ones coming from netlify went. Something else to watch out for is the upgrade of React to `16.4.0`, I'm testing it currently whether it works with client-core and all its dependencies. Other than that this should be working fully.

eos87 (approved):

PR #1003, opened 2025-12-18
[SDESK-7725] - Enhance CopyableTextBox

thecalcc:
### Purpose

Give more control over to the develop when using `CopyableTextBox`

[SDESK-7725]

[SDESK-7725]: https://sofab.atlassian.net/browse/SDESK-7725?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app-typescript/components/CopyableTextBox.tsx
+++ b/app-typescript/components/CopyableTextBox.tsx
@@ -5,15 +5,18 @@ import {InputWrapper} from './Form';
 import nextId from 'react-id-generator';
 import {IInputCommon} from './Form/InputWrapper';
 import {toasted} from './Toast';
+import {Position} from './ToastMessage';
 
 interface ICopyableTextBoxProps extends IInputCommon {
     value: string;
 
     /**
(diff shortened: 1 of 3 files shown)


eos87 (approved):

app-typescript/components/CopyableTextBox.tsx
```diff
@@ -52,6 +56,8 @@ export const CopyableTextBox: React.FC<ICopyableTextBoxProps> = (props) => {
             .catch(() => {
                 toasted.notify(gettext("Couldn't copy"), {
```
eos87, line 57: I'm wondering the scenarios for this to fail as the feedback is not very useful here.

thecalcc: it would tell you that text couldnt be copied so you'll look to copy it in another way or just manually move the text where you need it

PR #1004, opened 2025-12-19
[STT-1538] - Fix crash in month navigator label rendering

BrianMwangi21:
### Purpose
Fix a crash in the date picker month navigator that caused the “Start date” / “End date” fields in the Subscriber edit form to disappear when clicked.

### What has changed
- Updated DatePicker month navigator to resolve the selected month label safely  instead of relying on positional indexing. 

### Steps to test
1. Run the project and go to Settings → Subscribers.
2. Click Edit on any subscriber or add a new subscriber
3. Click Start/End date. Calendar should open and stay visible and there should be no console errors like can't access property 

[STT-1538](https://sofab.atlassian.net/browse/STT-1538)

[STT-1538]: https://sofab.atlassian.net/browse/STT-1538?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app-typescript/components/DatePicker.tsx
+++ b/app-typescript/components/DatePicker.tsx
@@ -337,6 +337,13 @@ export class DatePickerISO extends React.PureComponent<IDatePickerISO> {
     }
 }
 
+function getOptionLabel(options: Array<{id: string; label?: string}>, id: string | null | undefined): string {
+    if (id == null) {
+        return '';
+    }
+    return options.find((option) => option.id === id)?.label ?? id;
+}
+
 const MonthNavigator = ({
     value,
     options,
@@ -346,7 +353,7 @@ const MonthNavigator = ({
         kind="synchronous"
         value={[value]}
         getOptions={() => options.map((option) => ({value: option.id}))}
-        getLabel={(option) => options[parseInt(option, 10)].label}
+        getLabel={(optionId) => getOptionLabel(options, optionId)}
         getId={(option) => option}
         onChange={(selected) => {
             onChange(selected[0]);
@@ -355,7 +362,7 @@ const MonthNavigator = ({
         labelHidden
         valueTemplate={(item) => (
             <div className="sd-datepicker__navigator-value sd-datepicker__navigator-month">
-                <div>{options[parseInt(item, 10)].label}</div>
+                <div>{getOptionLabel(options, item)}</div>
                 <Icon name="chevron-down-thin" />
             </div>
         )}


app-typescript/components/DatePicker.tsx
```diff
@@ -337,6 +337,13 @@ export class DatePickerISO extends React.PureComponent<IDatePickerISO> {
     }
 }
 
+function getOptionLabel(options: Array<{id: string; label?: string}>, id: string | null | undefined): string {
+    if (id == null) {
+        return '';
+    }
+    return options.find((option) => option.id === id)?.label ?? id;
```
thecalcc: might gain some performance benefit if this is an object where key is `id` and `label` is value

eos87 (commented):
I'm not sure about this implementation. @BrianMwangi21 what is the root cause of the issue? is it incomplete or wrong data provided to the component?

eos87:
One more thing. If this fix is for STT we should probably not target `develop` as base branch. We should use the corresponding base branch according to the version they're using unless we will backport it.

BrianMwangi21:
@eos87 The root issue wasn't incomplete/wrong data. The field disappears immediately on click even before any user input and you'd get the following error message locally:
```
can't access property "label", options[parseInt(...)] is undefined
valueTemplate@webpack:////home/kabiru/Desktop/Work/superdesk-client-core/node_modules/superdesk-ui-framework/react/components/DatePicker.js?:272:54
TreeSelect.prototype.renderItemContent@webpack:////home/kabiru/Desktop/Work/superdesk-client-core/node_modules/superdesk-ui-framework/react/components/TreeSelect/TreeSelect.js?:551:31
TreeSelect.prototype/render/<@webpack:////home/kabiru/Desktop/Work/superdesk-client-core/node_modules/superdesk-ui-framework/react/components/TreeSelect/TreeSelect.js?:640:82
TreeSelect.prototype.render@webpack:////home/kabiru/Desktop/Work/superdesk-client-core/node_modules/superdesk-ui-framework/react/components/TreeSelect/TreeSelect.js?:625:34
finishClassComponent@webpack:////home/kabiru/Desktop/Work/superdesk-client-core/node_modules/react-dom/cjs/react-dom.development.js?:17039:3
```

The old code made an unsafe assumption:
`options[parseInt(item, 10)].label  // Assumes item is always a valid array index`

Hence when clicked
- PrimeReact passes a month ID
- Code does parseInt() to get array index
- If this returns undefined (for example due to min/max date filtering causing index mismatches) it cause the crash and the above error is thrown.

The fix now uses a safer ID-based lookup instead of array position, which works regardless of array structure.

---

Also, since the client-core on stt where the error happened was using release tag v5.1.2 of the ui-framework, I have changed the base branch to v5.1.

Edit: I changed to v5.1 and it included more files than those I changed. So I might close this one and base my work off of v5.1

BrianMwangi21:
Created new PR based off of v5.2 -> https://github.com/superdesk/superdesk-ui-framework/pull/1005

PR #1005, opened 2026-01-01
[STT-1538] - Fix crash in month navigator label rendering

BrianMwangi21:
### Purpose
Fix a crash in the date picker month navigator that caused the “Start date” / “End date” fields in the Subscriber edit form to disappear when clicked.

### What has changed
- Updated DatePicker month navigator to resolve the selected month label safely  instead of relying on positional indexing. 

### Steps to test
1. Run the project and go to Settings → Subscribers.
2. Click Edit on any subscriber or add a new subscriber
3. Click Start/End date. Calendar should open and stay visible and there should be no console errors like can't access property 

[STT-1538](https://sofab.atlassian.net/browse/STT-1538)

[STT-1538]: https://sofab.atlassian.net/browse/STT-1538?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app-typescript/components/DatePicker.tsx
+++ b/app-typescript/components/DatePicker.tsx
@@ -337,6 +337,13 @@ export class DatePickerISO extends React.PureComponent<IDatePickerISO> {
     }
 }
 
+function getOptionLabel(options: Array<{id: string; label?: string}>, id: string | null | undefined): string {
+    if (id == null) {
+        return '';
+    }
+    return options.find((option) => option.id === id)?.label ?? id;
+}
+
 const MonthNavigator = ({
     value,
     options,
@@ -346,7 +353,7 @@ const MonthNavigator = ({
         kind="synchronous"
         value={[value]}
         getOptions={() => options.map((option) => ({value: option.id}))}
-        getLabel={(option) => options[parseInt(option, 10)].label}
+        getLabel={(optionId) => getOptionLabel(options, optionId)}
         getId={(option) => option}
         onChange={(selected) => {
             onChange(selected[0]);
@@ -355,7 +362,7 @@ const MonthNavigator = ({
         labelHidden
         valueTemplate={(item) => (
             <div className="sd-datepicker__navigator-value sd-datepicker__navigator-month">
-                <div>{options[parseInt(item, 10)].label}</div>
+                <div>{getOptionLabel(options, item)}</div>
                 <Icon name="chevron-down-thin" />
             </div>
         )}
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "superdesk-ui-framework",
-    "version": "5.1.5",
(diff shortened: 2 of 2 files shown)


BrianMwangi21:
Closed the previous [PR](https://github.com/superdesk/superdesk-ui-framework/pull/1004) and opened this one based off of v5.1

@eos87 , to answer your question again on the root cause : 

The root issue wasn't incomplete/wrong data. The field disappears immediately on click even before any user input and you'd get the following error message locally:
```
can't access property "label", options[parseInt(...)] is undefined
valueTemplate@webpack:////home/kabiru/Desktop/Work/superdesk-client-core/node_modules/superdesk-ui-framework/react/components/DatePicker.js?:272:54
TreeSelect.prototype.renderItemContent@webpack:////home/kabiru/Desktop/Work/superdesk-client-core/node_modules/superdesk-ui-framework/react/components/TreeSelect/TreeSelect.js?:551:31
TreeSelect.prototype/render/<@webpack:////home/kabiru/Desktop/Work/superdesk-client-core/node_modules/superdesk-ui-framework/react/components/TreeSelect/TreeSelect.js?:640:82
TreeSelect.prototype.render@webpack:////home/kabiru/Desktop/Work/superdesk-client-core/node_modules/superdesk-ui-framework/react/components/TreeSelect/TreeSelect.js?:625:34
finishClassComponent@webpack:////home/kabiru/Desktop/Work/superdesk-client-core/node_modules/react-dom/cjs/react-dom.development.js?:17039:3
```

The old code made an unsafe assumption:
`options[parseInt(item, 10)].label  // Assumes item is always a valid array index`

Hence when clicked
- PrimeReact passes a month ID
- Code does parseInt() to get array index
- If this returns undefined (for example due to min/max date filtering causing index mismatches) it cause the crash and the above error is thrown.

The fix now uses a safer ID-based lookup instead of array position, which works regardless of array structure.

thecalcc (approved):

eos87:
> // Assumes item is always a valid array index

@BrianMwangi21 do you know why wouldn't it return a valid item? I mean, what could be the reason for it returning a wrong value?

BrianMwangi21:
@eos87 The reason it returns a wrong value is that `parseInt(item, 10)` converts PrimeReact's month ID to an array index, but when the month array is filtered by minEndDate (which we do in subscribers where we set the min end date dynamically where end date = selected start date + 1 day), that index no longer exists in the smaller array. So `options[parseInt(item, 10)]` becomes undefined instead of the expected month object and throws an error when we try and get the label using `.label`

eos87 (approved):
Looks good! @BrianMwangi21 please just bump the version number (patch) and it will get automatically pushed to npm.

PR #1016, opened 2026-02-12
Checkbox visibility, token separation & SCSS refactoring

fritzSF:
- Added a background color for the unchecked Checkbox state to improve visibility in overlay mode
- Refactored `_checkbox.scss` to use CSS custom properties instead of SCSS variables
- Extracted design tokens for Checkbox, CheckButton, RadioGroup, and RadioButtonGroup into dedicated files, separate from the core SCSS
- Updated SCSS imports in `_design-tokens.scss` to reflect the new file paths
- Fixed checkbox alignment in list items (selection indicator)
- Standardized icon font imports across example components
- Updated webpack configuration to use `icss` for CSS modules, resolving blank rendering on icon font documentation pages

[SDESK-7866]

[SDESK-7866]: https://sofab.atlassian.net/browse/SDESK-7866?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


webpack.config.js
```diff
@@ -13,6 +13,7 @@ const cssLoaderConfig = {
         url: {
             filter: (url) => !url.startsWith('/'),
         },
+        modules: 'icss',
```
thecalcc, line 16: i hope this doesn't break something in the projects importing ui framework, @fritzSF did you check if you link UI framework to e.g. client-core if that breaks something?

thecalcc (approved):

thecalcc:
As a general thought - I think it won't require too much time to handle replacement of old token names with new ones - if those aren't used a lot in outside projects using the IDE's built in find an replace with exact case and text match should allow for an easy transition. If that's the case we could remove the tokens with the old names

webpack.config.js
fritzSF: No, I did not check.

fritzSF:
> As a general thought - I think it won't require too much time to handle replacement of old token names with new ones - if those aren't used a lot in outside projects using the IDE's built in find an replace with exact case and text match should allow for an easy transition. If that's the case we could remove the tokens with the old names

That's the thing: a lot is used in Superdesk. It would need to be coordinated in both SD and the UI framework.

webpack.config.js
eos87: We should run E2E tests in this project too to avoid this things. We would need to set up the Github workflow to use the latest changes from the branch of the work in progress and run them. Thoughts?

package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "superdesk-ui-framework",
-    "version": "6.1.2",
+    "version": "6.1.3",
```
eos87, line 3: @fritzSF maybe we should not bump the version now if there are going to be more related changes. Also, if it's a breaking change, we should have to bump to v7.

eos87 (changes requested):

eos87 (approved):

PR #1039, opened 2026-07-24
Rework Dropdown to render menus via portals

fritzSF:
Replace the manual ReactDOM.render/unmountComponentAtNode placeholder
handling with createPortal, so menus and submenus participate in the
normal React tree and unmount cleanly.

- Extract IMenuElement and type `each` properly instead of `any`
- Portal submenus to document.body so an overflow-scrolled parent menu
  (maxHeight) cannot clip them
- Keep submenus open while the pointer travels between levels, with a
  short close delay and a hover context shared with ancestors
- Destroy popper instances on close instead of leaking them
- Treat submenus with no items as plain menu items
- Move the toggle's white-space rule onto a dedicated
  dropdown__toggle--default class so custom toggles are unaffected
- Document maxHeight and add a scrollable-with-submenu example

Also cap side-panel xx-large/xxx-large containers at 50vw/60vw.


fritzSF:
I gave Claude Code the review from Copilot. Here is what he says:
```
read Dropdown.tsx in full. Here's the honest assessment:

What Copilot got right: ensureDropdownContainer() is called at Dropdown.tsx:233, inside the JSX return — so yes, a DOM mutation happens during render. That violates the render-purity rule. As a code-review nit, it's legitimate.

What Copilot got wrong: "this can run multiple times and cause unexpected DOM mutations." The function is idempotent by construction — line 47 does getElementById first and only appends when the node is absent. Calling it 100 times produces exactly one container. There is no duplicate-node scenario, no leak, no observable bug.

Two facts that shrink the risk further:

This repo is on React 16.14.0 (react and react-dom both pinned). React 16 has no concurrent rendering at all, so the "discarded render leaves a stray node" scenario cannot happen here.
Even if it did: the stray node is a single 1×1px, position:absolute, empty, invisible div at 0,0. Worst case is one harmless element in <body>.
The one thing that keeps this from being a pure non-issue: tasks/webpack.prod.js:6 marks react/react-dom as externals, so the consumer (superdesk-client-core) supplies React. If a consumer is on React 18+ with StrictMode, the concurrent-rendering path becomes theoretically reachable — still with the same harmless outcome.

Worth noting: the pre-rework version (2e42288b^) created this container in a useEffect. So the fix is a return to the prior pattern, not a new invention.

If you want it fixed anyway (recommended, low risk)
Three lines, no behavior change:

const [container, setContainer] = React.useState<HTMLElement | null>(null);

// Create the shared portal container after mount so render stays side-effect free.
React.useLayoutEffect(() => {
    setContainer(ensureDropdownContainer());
}, []);
and at line 233:

{open && container != null && ReactDOM.createPortal(renderMenu(), container)}
Why this is safe: the mount-time useLayoutEffect runs and re-renders synchronously before paint, and the menu only ever opens on a user click — long after mount. So container is always non-null by the time open flips. No flicker, no delayed menu. Cost is one extra render per Dropdown on mount.

What I would not do: portal directly to document.body (the way DropdownItemWithSubmenu does at line 384) and delete the container entirely. It's tempting — the container serves no positioning purpose since popper uses strategy: 'fixed' — but it carries data-test-id="dropdown-overlay", added deliberately in [#897](https://github.com/fritzSF/superdesk-ui-framework/issues/897), and superdesk-client-core e2e tests likely select on it. Not worth the downstream breakage.

One unrelated nit while I was in there: React.useState(getNextZIndex) at line 65 also mutates module state (zIndex.tsx's counter) during render. Under StrictMode double-invoke it just burns a number. It's the standard React idiom and I'd leave it alone — mentioning it only so you're not surprised if Copilot flags it next.

No Dropdown spec file exists, so nothing in the test suite covers this either way.
```
I applied the change Claude suggested

package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "superdesk-ui-framework",
-    "version": "7.0.0-dev1",
+    "version": "7.0.1-dev1",
```
BrianMwangi21: @eos87 question, should the bump be this or `7.0.0-dev2` or `7.0.1` is there a pattern we should follow ? Just for future reference ?

app-typescript/components/Dropdown.tsx
```diff
@@ -15,266 +15,229 @@ interface IMenuItemRes extends IMenuItem {
...
+        );
+    }
+
+    function renderMenu() {
+        // Only constrain overflow when scrolling is requested. Submenus are portaled
+        // to document.body so they are not clipped by this overflow.
+        const menuStyle: React.CSSProperties = {
+            zIndex,
+            display: 'block',
+            ...(maxHeight != null ? {maxHeight, overflowY: 'auto' as const} : {}),
```
BrianMwangi21, line 182: Wondering if a header/footer would scroll away with the content as-is.

Would moving `overflowY: auto + maxHeight` onto the `.dropdown__menu-body <ul>` keep them pinned?

package.json
eos87: so after our discussion, give this will land in sd-develop, this should be `7.0.0-dev2`. @fritzSF can you fix it please?

app-typescript/components/Dropdown.tsx
fritzSF: > Wondering if a header/footer would scroll away with the content as-is.
Header and footer are not meant to scroll.

fritzSF: > Would moving overflowY: auto + maxHeight onto the .dropdown__menu-body <ul> keep them pinned?
@BrianMwangi21 I don't get the question. What do you mean by pinned?

package.json
fritzSF: done.

eos87 (approved):

